PR #121, opened 2026-07-16
PR-C: move ordinary income to the runtime profile (#7)

MattOrefice:
Audit item **#7 — income constant in a tracked file**. Third of the four sequenced PRs.

## The move

`ORDINARY_INCOME_2026` was a literal in `src/location_config.py` — tracked, in a now-public repo — and quoted again in three comments around it (`federal_marginal`, `federal_ltcg`, and the headroom block). It now lives in `private/personal_profile.json` (gitignored), read at runtime via `personal_profile.load_income_profile()`, mirroring `load_roth_profile` exactly: same shape, same graceful degradation.

- `personal_profile.example.json` ships a placeholder (`0`). **No real figure is committed** — verified the diff contains none, and no tracked file does.
- Demo uses `DEMO_ORDINARY_INCOME = 100_000.0` — a round, obviously-invented salary, chosen above the ceiling so demo exercises the interesting (exhausted-bracket) path.
- `LTCG_0_BRACKET_CEILING_SINGLE_2026` **stays** in the tracked config. It's tax law, not a personal figure.
- Your real value was migrated into the gitignored profile, so personal mode is unchanged.

## The part that wasn't just a constant move

Making income configurable forced a question the hardcoded value had hidden: **what should the page say when income is unknown?**

The tempting fallback is zero — and it's the one genuinely dangerous answer available. The 0% bracket is sized as `(ceiling − income)`, so a zero stand-in reports the **entire $48,350 ceiling as free headroom**, i.e. advises realizing gains tax-free that are in fact taxed at 15% + PA. So unknown income returns `None`, collapses the budget to zero, and the page says why. Same instinct as a missing Roth basis leaving the whole Roth locked: the absent-profile default is the one that cannot mislead. Two tests pin this specific wrong answer as impossible.

**The prose was also wrong-in-waiting.** It asserted the 0% bracket was "out of reach at this income" and "exhausted" as flat statements — true only while income sat above the ceiling, which a hardcoded figure *guaranteed* and a configurable one does not. It now states the exhausted, real-headroom, and not-configured cases separately. **With income unchanged the rendered page is identical to before.**

**One design fix my own test caught:** `capital_gains_headroom(register, ordinary_income=None)` was ambiguous — None meant both "not supplied, resolve it" and "unknown". A test asking for the unknown case was being silently answered from the real profile. Split with a sentinel; they're genuinely different states.

## Verification

- Full suite **demo**: `1293 passed`, 0 failures.
- Full suite **personal**: `17 failed, 1250 passed` — **byte-identical** to the pre-existing baseline (`diff` clean). No new failures.
- 20 new tests: synthetic-demo-value assertions, unknown-vs-zero degradation, malformed/null/zero/negative profile handling, headroom arithmetic at and around the ceiling, and `hasattr` checks that the tracked config exposes no income constant.
- **Rendered personal-mode Assumptions block** inspected directly: income templates from `private/personal_profile.json`, headroom and the exhausted branch render exactly as before.
- **Demo**: Asset Location is personal-only (`app.py:321`), so it stays out of the nav; the modules still import and resolve the synthetic value cleanly.
- Verified `private/personal_profile.json` is ignored and nothing from `private/` is staged.


MattOrefice:
**Self-correction before merge.** The first push of this branch contained a leak I'd introduced myself: a test asserting `DEMO_ORDINARY_INCOME != 89_000.0`. That bakes the real income into a *tracked* test file — precisely the exposure this PR removes, and `git grep`-able forever.

Rewritten to assert what the constant **is** (`== 100_000.0`) rather than what it isn't, plus a test that demo never falls through to a private profile even when one exists. The real value is now named nowhere in the tree. Branch force-pushed (unmerged, no shared history).

**One unavoidable residue:** the diff's removal line (`-ORDINARY_INCOME_2026: float = 89_000.0`) necessarily shows the value being deleted. It is not in the tree at HEAD — `git grep` is clean — but it is reachable via `git log -p` on this commit. That's the same accepted risk class as the CHANGELOG redaction in #118: front-page/tree exposure closed, deep-history reachable, no second `filter-repo` cycle. Flagging it rather than letting it pass silently.

PR #126, opened 2026-07-16
PR-1: de-silence the SAA guards before the sleeve split

MattOrefice:
Tests and test-support only. **No `src/` changes** — `SORT_ORDERS`, `_CELL_OFFSETS`, `_EQ_NAMES`, and all `src/` logic are untouched.

Every guard below is shown **failing under a deliberate mutation**. Mutation output is in the comment thread.

## The new guard — `tests/test_saa_sleeve_coverage.py`

`compute_household_allocation` derives the SAA sleeve set from the **securities join**, not from `asset_classes`. A sleeve whose `is_in_saa` carrier disappears does **not** become a zero-exposure row: the backfill at `household.py:244` iterates `saa_name_to_target`, which never contained it. The sleeve **vanishes from the frame entirely** — no row, no drift, no error — while its target still sits in the DB, and the function's only assertion (`:207`, dollar sums) still reconciles because the money relabels itself off-SAA.

- **(a)** on-SAA sleeve names in the returned frame == DB strategic names
- **(b)** their target weights sum to the DB's, within 1e-4

Both sides read live. A third test applies the regression and asserts **both guards break** — so they can't quietly go vacuous the way `test_sleeve_weights_sum_to_one` did.

**⚠️ One deviation, forced by the data.** The spec said read demo mode directly. `demo.db` carries `sleeve_category` and `is_in_saa` as **NULL on all 21 securities** — the household mapping is personal-mode only (CSV → `tracker.db`), and demo's schema is deliberately additive-and-inert. Reading demo.db alone leaves `saa_secs` empty, every position resolves off-SAA, and the guard fails on arrival for the wrong reason. So the securities frame is reconstructed from the two committed sources CI *does* have: **demo.db's real `asset_class_id`** (from `seed_securities.py`) + **the committed `securities_household.csv`'s real `sleeve_category`/`is_in_saa`**. That's the same shape personal mode assembles, from data in the repo. Nothing hardcodes a sleeve list or count.

## De-silencing

| Site | Was | Now |
|---|---|---|
| `test_attribution.py:860` | `pytest.skip` when the sleeve is absent — **exactly what the split does**, retiring the DRIP double-count tripwire silently | hard assert naming the sleeves present, refusing a re-skip |
| `test_factors.py:783` | `>= 1` — tolerated a dropped sleeve (the builder allows partial results by design) | exact against `_SLEEVES`, plus **three config↔DB ties that did not exist at all** |
| `test_asset_evaluation.py:80` | `sum(SLEEVE_WEIGHTS) ≈ 1.0` — vacuous | `_RAW_WEIGHTS` shape vs DB + **scale** vs the seed's normalizer |
| `test_harvest.py:154` | hand-listed tickers | derived from the committed CSV |
| `test_positioning.py:105` | hand-listed tickers | derived from CSV × DB Equity sleeves |
| `test_securities_mapping.py:277` | personal-only, never runs in CI | CI-runnable **structural** equivalent |

## Three things worth calling out

**`test_asset_evaluation.py:80` was worse than "vacuous".** Line 62 *defines* `SLEEVE_WEIGHTS = {k: v / _raw_total}`, so the sum is 1.0 by construction. Proof: I doubled all nine raw weights (0.98 → 1.96) and **the old assertion passed** — `sum == 1.0000000000` exactly. The re-pointed test pins scale against `_EXCASH_NORM`, which the normalization hides. It also trips on `Intl Developed 0.20 → 0.205` — the proposed split's arithmetic error.

**The CI equivalent for `securities_mapping` is structural, not another frozenset.** A literal set can't fail on a ticker it doesn't mention — the same defect being removed elsewhere. Instead: exactly one `is_in_saa` carrier per sleeve, none missing. **The duplicate case was unguarded and matters** — `household.py:153` `drop_duplicates("sleeve_category")` keeps whichever row comes first in the CSV and discards the rest, so a second carrier means the deploy buys an arbitrary ticker.

**I caught myself weakening a guard.** My first derivation for `test_positioning` used `sleeve_display_name`, which produced `['IEMG','SPHQ','VEA','VOO','VTV']` — **AVUV missing**, because `us_small_value` maps to the SAA sleeve "US Small Cap" and the display names don't round-trip. That would have been weaker than the literal it replaced. Now resolved by `asset_class_id`, the same join `household.py` uses; the derived set equals the old literal exactly and auto-extends.

## Verification

- Full suite **demo**: `1302 passed`, 0 failures (was 1294; +8).
- Full suite **personal**: `17 failed, 1259 passed` — **identical** to baseline (`diff` clean).
- **Headless render** (demo, port 8601): HTTP 200, no errors, and the guard reads the same frame the Household View renders — 9 on-SAA sleeves, targets summing to 1.0000000000.
- Working tree confirmed: **tests only**, no `src/`, no `demo.db`, no cache files staged.


MattOrefice:
## Mutation output — every guard shown failing

Each mutation was applied, the failure captured, then reverted. No mutation is committed.

### 1. New guard — sole `intl_developed` carrier loses `is_in_saa`

```
FFF
E  AssertionError: On-SAA sleeves in the returned frame do not match asset_classes.
E      missing from the frame (silently dropped): ['International Developed']
E      present but not a strategic target:        []

E  AssertionError: On-SAA target weights sum to 0.7959183673 but asset_classes
   sums to 1.0000000000 (delta 2.04e-01).
```

Both guards fail, and this **confirms the Part 1d trace**: the sleeve *vanishes* (not a 0% row), and targets silently fall to **79.59%**. The third test also fires — it detects the guards are unprotected.

### 2. `test_attribution.py` — sleeve renamed (simulating the split)

```
E  AssertionError: 'International Developed' is absent from the BF result, so this
   test's premise no longer holds. Sleeves present: ['Core Fixed Income', 'Emerging
   Markets', 'International Developed', 'Real Assets', 'TIPS', 'US Large Core',
   'US Large Quality', 'US Large Value', 'US Small Cap']. Do NOT re-skip this —
   re-point it at a single-ticker sleeve and update the VEA price lookup below, or
   the adj_close/DRIP double-count tripwire is gone.
```

Previously this path was `pytest.skip` — **green, tripwire gone.**

### 3. `test_factors.py` — a third sleeve added whose result never materializes

```
E  AssertionError: Expected one section entry per configured factor sleeve
   (3: ['developed_exus', 'intl_value_MUT', 'us']), got 2. With both regressions
   mocked to succeed, a shortfall means a sleeve was silently dropped — do NOT
   relax this back to >=.
E  assert 2 == 3

E  AssertionError: _SLEEVES names tickers absent from securities: ['AVIV'].
   A factor sleeve is regressing an instrument the portfolio does not define.
```

The old `>= 1` **would have passed** (2 ≥ 1).

### 4. `test_asset_evaluation.py` — the decisive one

**Mutation A — every raw weight doubled (0.98 → 1.96), proportions unchanged:**

```
raw total   = 1.9600  (was 0.98)
SLEEVE_WEIGHTS sum = 1.0000000000
OLD assertion abs(total-1.0) < 1e-9 -> True   <-- WOULD HAVE PASSED
```
```
E  AssertionError: _RAW_WEIGHTS sum to 1.9600000000, but seed_saa._EXCASH_NORM is
   0.98 — the raw weights are no longer the SAA's raw weights. Normalization hides
   this downstream, so nothing else would catch it.
```

**Mutation B — `Intl Developed` 0.20 → 0.205** (the proposed split's total):

```
E  AssertionError: _RAW_WEIGHTS sum to 0.9850000000, but seed_saa._EXCASH_NORM is 0.98
```

### 5. `test_harvest` / `test_positioning` — `IEMG → AVES` in the CSV

```
E  AssertionError: Missing REPLACEMENT_MAP entries: ['AVES'].
E  AssertionError: ETF_STYLE_BOX missing equity ETFs: ['AVES'].
```

Both hand-lists would have passed — neither mentions `AVES`.

### 6. CI structural guards

**Two carriers** (`DFAE` flipped to `is_in_saa=true`):
```
E  AssertionError: Multiple is_in_saa tickers share a sleeve_category:
E    symbol  sleeve_category
E      IEMG emerging_markets
E      DFAE emerging_markets
E    household.py:153 drop_duplicates() would keep whichever comes first in the CSV
E    and silently discard the rest.
```

**Zero carriers** (`IEMG` flipped to `false`):
```
E  AssertionError: On-SAA sleeves in the returned frame do not match asset_classes.
E      missing from the frame (silently dropped): ['Emerging Markets']
```

All mutations reverted; `git diff --quiet -- data/seed/securities_household.csv` clean; suite green.

PR #127, opened 2026-07-16
PR-1b: arm the tripwires for the International Developed rename

MattOrefice:
One concern: make the rename's failure loud **before** the rename. No other `src/` changes; `SORT_ORDERS`, `_CELL_OFFSETS`, `_EQ_NAMES` untouched.

## 1. The two silent fallbacks

`pages/1_SAA.py:91` and `pages/2_Performance.py:315` resolved the sleeve target with a `0.20/0.98` default.

**The no-op proof came first, as instructed — and it surfaced why this matters.** The lookup resolves in **both** demo and personal today; neither default fires. So these are tripwires, not a live bug.

But the default is `0.20408163265306123` — **bit-identical to the live DB target**:

```
1_SAA.py:91      sleeve present in sub_classes : True
                 resolved value                : 0.20408163265306123
                 default would be              : 0.20408163265306123
                 default FIRED today?          : False
!! resolved == default : True
```

A silent fallback was therefore **indistinguishable from a correct lookup by value alone** — only membership distinguishes them. The rename would have templated a stale weight into the investment-thesis prose and the non-US-equity share with no error and no visible change. Both now raise, naming the sleeve and listing what's present.

**Mutation proof** — renamed the sleeve in `demo.db` and rendered the real pages:

```
pages/1_SAA.py:99 in <module>
ValueError: SAA sleeve 'International Developed' is absent from asset_classes, but this
page templates its target weight into the investment-thesis prose below. Sleeves present:
['Core Fixed Income', 'Emerging Markets', 'International Core', 'Real Assets', 'TIPS',
'US Large Core', 'US Large Quality', 'US Large Value', 'US Small Cap']. ...

pages/2_Performance.py:320 in <module>
ValueError: SAA sleeve 'International Developed' is absent from asset_classes, but this
page templates the non-US equity share into its performance prose. ...
```

`demo.db` restored **byte-identical** afterward (sha256 `6219338e…` before and after).

**Scope note:** there are **six** such fallbacks, not two — the sibling lookups (Equity, Emerging Markets, Real Assets, TIPS, Core Fixed Income) in the same blocks carry the identical defect. They're untouched: the rename doesn't affect them, and you scoped this to one concern. Worth a follow-up.

## 2. The personal-mode instrument

Adds the PR-1 invariant to the personal suite via `tests/test_household_aggregation.py`, reading `tracker.db` through the real path — no CSV reconstruction, real securities table, real look-through compositions, real positions.

**Mutation proof** (carrier dropped in memory; `tracker.db` never written):

```
E  AssertionError: On-SAA sleeves in the real household allocation do not match asset_classes.
E      missing from the frame (silently dropped): ['International Developed']
E      present but not a strategic target:        []

E  AssertionError: On-SAA target weights sum to 0.7959183673 but asset_classes sums to
   1.0000000000 (delta 2.04e-01).
```

It skips in CI by the file-presence gate this module already uses — verified: with `HOLDINGS_CSV=None`, `Skipped - Holdings CSV not present (personal-mode only)`.

## Verification

- **Personal: `17 failed, 1262 passed`** — failures **byte-identical** to baseline (`diff` clean). New guards: 3 passed.
- **Demo: `1305`, not 1302.** The +3 is the new personal guards, which run on this machine because they're gated by *file presence*, not `TRACKER_MODE` — `_load_fixtures()` reads `data/tracker.db` directly, which is this module's existing design. `data/uploads/*` and `tracker.db` are both gitignored, so **CI collects neither and the CI count stays 1302**. CI on this PR is the check.


MattOrefice:
**CI reconciles the count question exactly.**

| | passed | skipped |
|---|---|---|
| PR-1 (#126) baseline | 1222 | 80 |
| PR-1b (this) | 1222 | **83** |

**+3 skipped, +0 passed** — the three new personal guards skip in CI as designed, and the two tripwires change no CI behavior at all.

This also reconciles the "1302 vs 1305" discrepancy: **1302 was never CI's number.** CI collects 1305 and passes 1222 (83 personal-gated tests skip). Locally, `TRACKER_MODE=demo` still runs those 83 because `data/tracker.db` and the positions CSV exist on the dev machine and several personal test modules gate on *file presence*, not on `TRACKER_MODE` — so local demo reports 1305 passed / 0 skipped. Both numbers are correct for what they measure; the +3 is local-only.

PR #128, opened 2026-07-16
PR-1c: disarm the remaining hardcoded-default weight lookups

MattOrefice:
Same defect as #127 at every remaining site. Not a PR-2 blocker — none of these names or weights change under the split.

## Census — my #127 number was wrong

**15 sites: 2 armed + 13 remaining.** Not 6.

| File | Sites |
|---|---|
| `1_SAA.py` | :90 Equity · :99 Intl Dev *(armed #127)* · :108 EM · :109 Real Assets · :110 TIPS · :111 Core FI — **6** |
| `2_Performance.py` | :314 Equity · :320 Intl Dev *(armed #127)* · :328 EM · :811 US Large Value · :812 US Small Cap · :812 EM · :813 Real Assets · :813 TIPS · :814 Equity — **9** |

Your arithmetic caught it exactly. My "six" counted `1_SAA.py`'s block alone (6 lookups, 1 armed → 5 siblings) but I reported it as a repo-wide total against "2 armed," which spans two files. I'd missed `2_Performance.py:811-814` entirely — a prose caption with six more lookups.

## No-op proven first, per site, both modes

All 13 keys present; **no default fires**; and **every default is bit-identical to its live value** — `Equity 0.79591836734693877`, `EM 0.09183673469387756`, `TIPS 0.04081632653061225`, etc. A fallback could never be caught by value, only by membership.

Each site raises when its key is renamed in `demo.db`, verified by rendering the real pages:

```
pages/1_SAA.py:113 in <module>
ValueError: SAA parent 'Equity' is absent from asset_classes, but this page templates
its target weight into the investment-thesis prose below. Present: [...]
```

All 7 distinct keys checked. `demo.db` restored **byte-identical** (`6219338e…`).

Incidental: for `Core Fixed Income`, `2_Performance.py` raises a *different* guard first — `Sleeve(s) ['Core Fixed Income RENAMED'] carry benchmark weight but have no _SLEEVE_BENCHMARKS mapping`. Still loud, just an earlier tripwire.

## The finding that reframes #127

**The defaults were bit-identical BY TEST ENFORCEMENT, not by accident.** Two CI tests existed to extract the literals from page source and assert they matched the DB. Their docstrings note a *prior* version kept a separate copy and was deliberately rewritten to read the source instead.

So this was a maintained safety net — which means the fallback's silence was **guaranteed by design**, not merely likely. I described it in #127 as though it happened to line up. It didn't; it was kept that way on purpose. That makes the two tests the other half of the design, and they had to be retired with it.

## Inverted, not deleted

Per your call. Same extraction, asserted **empty**:

- `test_prose_methodology_weight_defaults_match_db` → **`test_prose_methodology_has_no_hardcoded_weight_defaults`**
- `test_saa_thesis_sleeve_weights_match_db` → **`test_saa_thesis_has_no_hardcoded_weight_defaults`**

Each also asserts its page still calls the raising accessor, so the guard can't pass vacuously against a page that resolves nothing.

**Mutation proof — all three fire:**

```
# reintroduce _saa_sleeves.get("TIPS", 0.04 / 0.98)
E  AssertionError: pages/2_Performance.py resolves SAA weights with hardcoded fallback
   defaults again: [('TIPS', '0.04 / 0.98')].

# reintroduce next((... == "TIPS"), 0.04 / 0.98)
E  AssertionError: pages/1_SAA.py resolves thesis weights with hardcoded fallback
   defaults again: [('TIPS', '0.04 / 0.98')].

# remove the accessor
E  AssertionError: pages/1_SAA.py no longer calls _require_weight() — the raising
   accessor was removed, so this guard would pass vacuously.
```

## Verification

- **Personal: `17 failed, 1262 passed`** — byte-identical to baseline. The 3 guards from #126/#127 pass; `test_saa_sleeve_coverage` 5 pass.
- **Local demo: 1305 passed** (the file-presence-gated personal tests run locally, as reconciled in #127).
- **CI: 1:1 replacement, so 1222 / 83 expected.** Reporting the actual number below before merge rather than assuming — it hasn't held twice now.


MattOrefice:
**CI reconciled — the 1:1 replacement held.**

| | passed | skipped |
|---|---|---|
| #127 baseline | 1222 | 83 |
| PR-1c (this) | **1222** | **83** |

Unchanged, as predicted: two tests replaced by two tests, and the 13 disarmed lookups are a no-op against live data (every key resolves; no default was ever firing). Reported from the run log rather than assumed.

PR #129, opened 2026-07-17
PR-1d: arm the DB-read-key sites (benchmark map drift, SLEEVE_META)

MattOrefice:
Same defect as #127/#128, different shape. #128's census signature was `.get(<literal>, <literal>)` and **structurally could not find these** — here the KEY comes from the DB, so there's no literal to grep for.

## The wrong number, on the record

Renamed `International Developed` → `International Core` in `demo.db` and called the blend directly (bypassing the page, so `attribution.py` couldn't raise first):

| | resolved weights | blend 2025-05-01 → 2026-06-30 |
|---|---|---|
| Baseline | **1.000000** | **+32.03%** |
| After rename | **0.795918** | **+33.23%** |

The stale key resolved to `0.0`. EFA was fetched and weighted **zero**. The 20.41% international leg vanished, the survivors **implicitly renormalized to $1** (as `benchmarks.py:200-208` documents), and the blend returned a plausible number with **`benchmark_gaps == []` and no error**.

**120 bps too high.** Dropping the laggard sleeve and rescaling the rest *inflates* the benchmark the portfolio is judged against — so the portfolio would have read 120bps **worse** against a blend that had silently deleted its international sleeve.

## Why `attribution.py:333` doesn't catch it

It checks **DB→map**: a live sleeve carrying weight with no benchmark column. A rename is **map→DB** — the stale key resolves to zero and *nothing downstream is missing*, so its check passes cleanly. Reused its machinery, widened to set equality in both directions, naming both sides:

```
AssertionError: _SLEEVE_BENCHMARKS and asset_classes disagree — benchmark configuration drift.
  in _SLEEVE_BENCHMARKS but NOT in the DB (resolves to weight 0.0 and silently drops its leg): ['International Developed']
  in the DB but NOT in _SLEEVE_BENCHMARKS (never priced): ['International Core']
```

`Cash / SPAXX` is expected in the map at target 0.0 (Phase 38a), so the comparison is against every sub-class rather than only the weighted ones.

## `seed_investment_theses.py:192`

`SLEEVE_META.get(sleeve_name, {})` with the name read from the DB. Verified: `SLEEVE_META.get("International Core", {})` → `{}`. Seeds a thesis with **no exit conditions, no invalidation conditions, no themes** — and `_conviction` falls through to its weight-based default, which returns *the same answer* for a >15% sleeve, so nothing even looks different. It also **orphans**: the title f-strings the new name, the already-seeded guard misses, and a second thesis is written while the old one stays behind. Now raises.

## No-op proven first, both sites, both modes

`_SLEEVE_BENCHMARKS` keys == DB sub-classes (10 == 10); `_SLEEVE_HOLDINGS` likewise; every DB sleeve has a `SLEEVE_META` entry. Neither default is reachable today.

## Census — reported, not fixed

An AST walk (regex is what made #128 blind here) found **105** `.get(<variable>, <default>)` call sites; filtered to **21** where a *module-level hardcoded map* is keyed by a *DB-read value*:

| File:line | Map | Key | Default |
|---|---|---|---|
| `reports.py:119,120` | `SLEEVE_HOLDING_TICKER` / `SLEEVE_BENCH_TICKER` | `sleeve` | echoes the key |
| `reports.py:763,764` | same | `sleeve` | `'—'` |
| `household.py:532` | `_SLEEVE_DISPLAY_NAMES` | `sleeve_key` | title-cased key |
| `1_SAA.py:33` | `_DISPLAY_NAMES` | `name` | echoes the key |
| `1_SAA.py:155,194` | `PARENT_COLORS` | `dname` | `'#888888'` |
| `location_config.py:248,256` | `SLEEVE_PRIORITY_BY_ACCOUNT_TYPE` | `account_type` | `{}` |
| `positioning.py:343` | `ETF_DURATION` | `ticker` | `0.0` |
| `drip.py:62` | `_PAYMENT_DATE_OVERRIDES` | `ticker` | offset const |
| `10_Trade_Log.py:40` | `STATUS_COLOR` | `status` | `'#888'` |
| `household_accounts.py:56` | `_TAX_TO_TYPE` | `tax_treat` | `'other'` |
| `seed_investment_theses.py:166,192` | `SLEEVE_META` | `sleeve_name` | `{}` ← fixed here |
| + `11_Capital_Deployment.py:183`, `5_Asset_Evaluation.py:1016` | | | |

**21 scopes the degradation audit as real but bounded** — closer to "small" than "outranks the restructure". Most render a visible placeholder (`'—'`, grey `#888`) rather than a plausible wrong number, which is a materially milder failure than the blend's silent rescale.

**A third shape exists that even this census can't see:** `benchmarks.py:181` is *inverted* — DB-built map, hardcoded key. Neither #128's signature nor this one finds it. It was found by tracing, not grepping.

## Verification

- **Personal: `17 failed, 1262 passed`** — byte-identical to baseline; guards from #126/#127 pass (3 + 5).
- **Local demo: 1305 passed.**
- CI number read from the run log below, not assumed.


MattOrefice:
**CI read from the run log: `1222 passed, 83 skipped`** — identical to the #128 baseline. No test behavior change: both guards are unreachable against live data (proven as a no-op per site, both modes, before the fix).

PR #135, opened 2026-07-22
CI baseline check (no-op) — do not merge

MattOrefice:
Empty commit off main to test whether the attribution reconciliation failures on PR #134's CI run reproduce on unmodified main under the same UTC date / fresh price data. Will be closed and deleted after comparison.


MattOrefice:
Baseline confirmed: unmodified main passed CI while #134's original demo.db failed, isolating the regression to leaked price-cache rows (fixed in 1c0fcc0). Closing.

Issue #247, opened 2026-08-18
Record the collection total with the split, and re-measure a baseline after the final commit


MattOrefice:
**REWRITTEN. Both original findings are withdrawn. One rule survives, and it is not the one this issue was filed for.**

The original title claimed a per-file pytest count is not a subset of the suite count. That is false in this repo, and the second finding — a baseline that did not reconcile to its own collection — had a completely different cause than the one recorded.

---

## Finding 1: WITHDRAWN. The per-file property does not exist.

Measured today at `972b137`, `tests/test_reports.py` in isolation:

```
guards down   53 passed
guards up     53 passed
```

The seven failures this issue was filed on are **not reproducible in either configuration**. And the general claim was swept mechanically — every test file run alone, its failure set compared against its contribution to the suite failure set, in **both** directions:

```
104 test files
0 divergent
```

Proof the sweep was not blind — it found exactly the 15 known baseline failures, correctly distributed:

```
test_asof_banner_render.py        2F alone / 2F in suite
test_factor_profile_render.py     4F alone / 4F in suite
test_performance_render.py        5F alone / 5F in suite
test_trade_log_render.py          3F alone / 3F in suite
test_securities_mapping.py        1F alone / 1F in suite
test_reports.py                   0F alone / 0F in suite
```

Including the worse direction — *passes alone but fails in the suite*, which a per-file check would clear — **zero instances**.

### The claim was never established

"An ordering or fixture dependency" was **inferred from two numbers differing and written as a cause.** No fixture was identified, no ordering was traced, no mechanism was shown. It was reviewed and approved as a finding on the strength of the inference. That is the same failure this repo has now recorded three times: a conclusion whose reasoning is plausible and whose contents were never checked against the code.

What actually caused the seven failures is **unknown and no longer knowable** — they occurred mid-session on the same tree, and the only plausible carrier is `data/tracker.db`, whose state at that moment is gone. Recording "unknown" rather than inventing a second mechanism to replace the first.

## Finding 2: KEPT, with its real cause. Not a pytest category.

The original explanation was "two tests in a category the one-line tail carries separately — most likely xfail/xpass or error". **Wrong on both counts.**

The saved output from that run is intact and its tail is complete:

```
15 failed, 1766 passed, 27 skipped, 40 deselected, 23 warnings in 313.36s
```

[...]

PR #252, opened 2026-08-18
Docs: record a baseline's collection total and re-measure after the final commit

MattOrefice:
Docs only — `CLAUDE.md`, +11 lines, nothing else. The evidence is a sweep, not a mutation proof.

## What the rule is

> **Record the collection total alongside the split, and confirm a recorded baseline belongs to the tree it names — re-measure after the final commit, not before it.**

`failed + passed + skipped` must equal the selected count from `pytest --collect-only -q | tail -1`. If it does not, the figure is incomplete and every delta computed from it inherits the gap.

It sits beside the #238 baseline discipline, which says quote a count with its harness and deliberately writes no count. That section did not say how to tell whether a recorded count belongs to the commit it names. It does now.

## Why, measured

At #246 the recorded baseline was `15 failed, 1766 passed, 27 skipped` — **1808 accounted for against 1810 collected** at `94fdb1e`.

The original explanation was a pytest summary category (xfail/xpass/error). **Wrong.** The saved output's tail is complete, and `grep -rn "xfail\|xpass"` over `tests/`, `src/` and `pytest.ini` returns **zero hits** — no xfail exists anywhere in this repo.

The actual cause:

```
full_pct.txt written at        18:31:42
collection at 94fdb1e          1810/1850 selected
15 + 1766 + 27               = 1808         +2 = 1810
```

**The two tests did not exist when the run happened.** The suite was measured mid-session; the conditional-caption work was approved afterwards and added `test_no_percentile_caption_references_a_raw_window_label` and `test_every_wired_panel_derives_both_scope_phrases` before the commit. A number measured on one tree, recorded as belonging to another.

**The reconciliation check worked exactly as designed** — it detected that a recorded figure did not describe the commit it named. Only the explanation was wrong.

### The exposure is structural, not incidental

This workstream takes its suite measurement after implementation and **before the final review round**. That ordering is sensible: the suite takes five minutes and the number is wanted for the PR body. But every review round that adds a test invalidates it, and reviews here routinely add tests — this PR's parent session added two, the one before that added two more.

So the recorded baseline is *systematically* at risk of describing an intermediate tree, and the arithmetic is the only thing that notices. That is why the rule belongs in the file every session reads first rather than in a note.

Applied to the current baseline, which reconciles: `15 + 1775 + 27 = 1817 = collected` at `972b137`.

## The omission is deliberate

**#247's per-file rule is not in this diff, and should not be.** That issue claimed a per-file pytest count is not a subset of the suite count, on the strength of `tests/test_reports.py` failing 7 alone and 0 in the full run.

Swept mechanically at `972b137` — all 104 test files run in isolation, each file's failure set compared against its contribution to the suite failure set, **in both directions**:

```
104 test files
0 divergent
```

Verified non-blind: the sweep found exactly the 15 baseline failures, correctly distributed (2 asof-banner, 4 factor-profile, 5 performance, 3 trade-log, 1 securities-mapping, 0 test_reports). And `tests/test_reports.py` now passes **53/53 alone in either guard state** — the seven failures are not reproducible in any configuration.

Including the worse direction — *passes alone but fails in the suite*, which a per-file check would clear — **zero instances**.

So the claim is false. **Writing a false rule into the file every session reads first is precisely what #238 exists to prevent**, and a rule there propagates into the premises of every future session, so a wrong one is worse than a missing one. The omission is the point, not an oversight.

### The claim was never established

"An ordering or fixture dependency" was inferred from two numbers differing and written as a cause — no fixture identified, no ordering traced, no mechanism shown. What actually caused those seven failures is **unknown and no longer knowable**: they occurred mid-session on the same tree, and the only plausible carrier is `data/tracker.db`, whose state then is gone. Recorded as unknown rather than replaced with a second invented mechanism.

## The positive result, recorded separately

**Per-file runs are valid in this repo today** — 0 of 104 divergent. A fast per-file run during development is trustworthy and needs no re-validation against the suite.

**Caveat, and a real limit: isolation is not ordering.** Running a file alone tests whether it is self-sufficient; it does not test whether two files interact *within* the suite. A pair that misbehaves only in a specific relative order passes both the isolated run and the ordinary full run, because the ordinary run has one fixed order and both measurements use it. Detecting that needs shuffled runs, which this sweep did not do.

That is a property of the repo rather than a rule for the reader, so it went to the notes rather than to `CLAUDE.md`.

## Ladder

- **No mutation proof** — this is prose. The sweep is the evidence, and it is reproducible: 104 files, both directions, non-blindness verified against the known failure set.

  That reproducibility is the point, and it is why a sweep is the right evidence for *this* PR specifically. A mutation matrix is a claim about tests that only its author ran; a sweep is a procedure anyone can repeat and get the same answer from. **For a change that withdraws a previously-approved finding, checkable-by-someone-who-trusts-neither-of-us is the property that matters** — the reader should not have to take either the original claim or its retraction on authority.
- **`tests/test_page_imports_resolve.py`: 17 passed** — the static guard that answers "do the page imports resolve at HEAD" in 0.27s. No other test touches `CLAUDE.md`.
- **Docs only**: `git status --porcelain` shows `M CLAUDE.md` and nothing else.
- **demo.db** drifted `prices +328` under the guards-down sweep and was restored — per-table fingerprint IDENTICAL across all 13 tables. Guards cycled and `verify` → no drift.

## Also in this change set

**#247 rewritten** — retitled to the surviving rule, finding 1 withdrawn with the sweep as evidence and an explicit statement that the claim was never established, finding 2 kept with its real cause.

**Notes corrected**, since both carried the false rule: the index line now points at the measured property, and `local-test-baseline` had two wrong sentences — the xfail explanation and the per-file prohibition — both replaced.


MattOrefice:
## How this actually merged, since the record is otherwise misleading

GitHub marked this MERGED at 03:22:45 with **no merge commit of its own** and with its CI run (`32091829873`, SHA `9da8855`) never having gone green — it was hung on the WeasyPrint apt step and has now been cancelled. That combination reads like a merge that bypassed verification. It was not, but the reason is not visible from this page.

**What happened:** the branch for #256 was created from `docs/baseline-provenance-rule` instead of from `main` — an unforced error, not a plan. So `9da8855` was already an ancestor of #256's branch, #256's CI ran on `1936c84` (which *contains* this commit's CLAUDE.md change), and merging #256 made `9da8855` reachable from main. GitHub auto-closed this PR as merged.

**Verification status, stated precisely:** the tree that landed is `c3b6f0be`, which is byte-identical to `1936c84`'s tree, and `1936c84` has a **green CI run**. So the merged state was CI-verified. What never got its own green is the intermediate commit `9da8855` — which is the ordinary situation for any non-final commit in a multi-commit PR, and is not what the by-SHA rule is about. The rule asks that the tip that lands be green, and it was.

**What is genuinely wrong:** #256's body describes a workflow-only change and its diff includes this CLAUDE.md change, which the body never mentions. A reader of #256 would not learn that the baseline-provenance rule landed there. That is corrected by a comment on #256 pointing here.

The content is unaffected — the rule is on main and reads as intended.

Issue #255, opened 2026-08-18
Should CI verify that the PDF builds? A separate decision from removing the apt step


MattOrefice:
Filed as a separate decision, deliberately not folded into #253.

`tests/test_market_data_immutability.py::test_pdf_build_writes_no_tracked_data` is the only test that reaches the PDF render:

```python
try:
    out = generate_quarterly_report_bytes("2025-05-01", "2026-06-30", is_demo=True)
    assert isinstance(out, bytes) and out, "PDF build returned no bytes"
except Exception:
    # A build failure is a different defect; the guarantee under test is
    # only that no tracked file was written along the way.
    pass
_assert_no_attempts(write_trap, "building the quarterly PDF")
```

**The swallow is deliberate and correct for what the test asserts.** Its guarantee is that no tracked file was written during a PDF build, and that holds whether the build succeeds or raises. Nothing about it is a defect.

The consequence, though, is that **CI does not verify that the PDF builds at all** — and it did not before #253 either. So the question is whether it should.

## Why this must not be decided by dependency convenience

While removing the hanging apt step in #253, the observation "nothing asserts on the PDF build" was the *argument for* removing it. The reverse move — tightening this test so it asserts the build succeeds — would give that apt step a purpose and justify re-adding it.

**That is backwards, and it is the reason this is a separate ticket.** Changing what a test asserts in order to justify a dependency inverts the relationship: the test should assert what the product must guarantee, and the dependency follows. Deciding PDF coverage because it would rationalise an apt step is choosing the requirement to fit the infrastructure.

So: if CI should cover the PDF build, that is because a broken PDF is a defect worth catching in CI — a claim about the product. Decide it on those terms, and if the answer is yes, the system libs come back as a *consequence*.

## What it would cost

- A new test (or a second assertion in a differently-named one) that calls `generate_quarterly_report_bytes` and requires bytes, with no swallow.
- The Pango/HarfBuzz system libs back in CI — which reintroduces the hang risk from #253, now bounded by the `timeout-minutes` that PR added, so a stall fails in 5 minutes instead of hanging.
- A decision on which package list is correct, since CI's and `packages.txt`'s disagree by two entries (tracked separately).

## Arguments on each side, briefly

**For covering it:** the PDF is a user-facing deliverable, it has had real defects (#193 basis, #204 capture, #244 fabricated stance), and a build that raises would currently reach a user before it reached a test.

[...]

Issue #254, opened 2026-08-18
packages.txt and CI's WeasyPrint lib list diverged by two packages — which one is right is untested


MattOrefice:
Found while removing the CI apt step (#253). **The divergence survives that removal** — it is a fact about the deployed environment, not about CI.

```
packages.txt  (Streamlit Cloud runtime)     libpango-1.0-0  libpangoft2-1.0-0  libharfbuzz0b        = 3
.github/workflows/ci.yml (before removal)   the same 3  +  libgobject-2.0-0  libffi-dev            = 5
```

CI installed **two packages the deployed environment does not have.** Whichever list is right, they disagreed, and only one of them describes where the PDF is actually generated for a user.

## Which direction is wrong is not established

Two readings, and I did not test either:

- **CI over-installed.** `libgobject-2.0-0` is part of glib2 and is a transitive dependency of `libpango-1.0-0`, so it is almost certainly present on Cloud regardless of whether `packages.txt` names it. `libffi-dev` is a *build* dependency (dev headers for cffi) and a runtime environment does not need it if wheels are prebuilt. Under this reading the divergence is harmless and `packages.txt` is correct.
- **Cloud under-installs.** If WeasyPrint genuinely needs either at import or render time, the PDF path could fail in production while CI passed — and CI passing was never evidence, because the one test that reaches the render swallows the exception (see #253).

The first reading is more likely. It is still a reading.

## How to settle it

The check is on Cloud, not here: generate a quarterly PDF from the deployed demo app and confirm it produces bytes. If it does, `packages.txt` is sufficient and CI's extra two were noise. If it fails on a missing shared library, `packages.txt` needs the additions and the deployed PDF path has been broken without anything reporting it.

That is a live-environment check, so it needs the owner rather than a test.

## Why it is worth recording rather than dropping

Two lists of system dependencies for the same library, maintained separately, disagreeing by two entries, with **no test that would notice** — the PDF test's swallow means neither list is validated by CI. So the divergence could have grown in either direction indefinitely. Even if the answer is "harmless", the mechanism that let them drift apart is worth a note in `packages.txt` pointing at the CI workflow, or vice versa.

Related: #253 removed the CI step (it guarded nothing there); the runtime requirement on Cloud is unaffected and unverified.

MattOrefice:
**The three-versus-five framing is wrong, not merely incomplete — and both candidate lists miss a required library.**

[...]

PR #256, opened 2026-08-18
CI cannot hang indefinitely, and the step that hung guarded nothing

MattOrefice:
Two changes answering two different questions. **A timeout makes a hang legible; it does not make green obtainable.** Only removing the dependency does that, and both are needed.

## 1. Timeouts — legibility

The workflow had **no timeout anywhere**, so the job inherited GitHub's 360-minute default. A stalled step meant up to six hours of neither pass nor fail.

```
job                              timeout-minutes: 15   (observed normal 3-4 min)
Set up Python                    timeout-minutes: 5
Install Python dependencies      timeout-minutes: 5    (observed ~30s)
Run fast tests                   timeout-minutes: 10   (observed ~2.5 min)
```

The test step gets more room deliberately: **a slow test run and a hung one deserve different verdicts**, and 10 against 2.5 leaves a runner room to be slow without letting a hang sit.

A failure is a signal; a hang is the absence of one, and an absence invites *"it would have passed"* — which is the substitution `CI green BY SHA` exists to forbid. **This half stays regardless of what happens to any individual step.**

## 2. The apt step goes — obtainability

It hung on `apt-get update -qq` **twice on the same SHA**, 25 and 17 minutes, against a measured **6-second** normal:

```
run 32085457063, step 4:  00:41:28 -> 00:41:34
```

On a docs-only diff (#252), with `-qq` suppressing the mirror output that would have said where it was waiting.

**It guarded nothing here, and that is the entire argument.** `weasyprint` is imported function-locally at `src/reports.py:236`, so importing `src.reports` — which `tests/test_imports.py` does for every module — never touches the system libs. The only test that reaches the render:

```python
try:
    out = generate_quarterly_report_bytes(...)
    assert isinstance(out, bytes) and out
except Exception:
    # A build failure is a different defect; the guarantee under test is
    # only that no tracked file was written along the way.
    pass
```

The swallow is deliberate and correct for what that test asserts. The consequence is that it **passes identically with or without the libs.** So an unbounded hang risk was being carried for a step no assertion depended on.

### Not dropped silently

The step is replaced by a comment recording why it left, that **the libs are still required at runtime on Streamlit Cloud via `packages.txt`**, and the two things a future re-adder needs to know (the lists diverged; PDF coverage is a separate decision). Without that, someone re-adds it and the hang risk returns with no record of why it went.

## Not tightening the PDF test

Making that test assert the build *succeeds* would give the apt step a purpose and justify re-adding it. **That is backwards.** A test should assert what the product must guarantee, and the dependency follows — choosing the requirement to fit the infrastructure inverts it. Filed as **#255** with both sides argued, including the strongest argument against: the interesting failures live in the context builders and the template, and those are fully reachable by patching `_render_pdf` and inspecting the HTML, with no WeasyPrint at all.

## Retry is not the companion

Buying flakiness tolerance for a step that should not gate the merge is the wrong trade, and a retry would mask a genuine dependency break that nothing detects anyway.

## Filed

**#254** — `packages.txt` lists 3 libs, CI installed 5 (`+libgobject-2.0-0`, `+libffi-dev`). **The divergence is about the deployed environment and survives this removal.** Which list is right is untested: CI over-installing is the likelier reading (glib is transitive, `libffi-dev` is build-only), but the check is on Cloud, not here.

**#255** — should CI verify the PDF builds at all.

## Ladder

- **No mutation matrix** — this is a workflow change, and the honest proof is the run itself. **This branch's CI run is the test**, including for the one real risk: `libffi-dev` was a build dependency for cffi, so if `pip install` now needs a source build it fails here rather than in a later PR.

  **Settled by the run: `Install Python dependencies` took 30s** (03:17:50 -> 03:18:20), its usual time, so cffi resolved from a wheel and no source build was needed. That counts as evidence rather than luck precisely because it was flagged as the risk beforehand — an unflagged green would have been the same 30 seconds and would have proved nothing about a question nobody asked.
- **YAML validated** with `yaml.safe_load`: job timeout 15, 9 steps (was 10), apt step absent, timeouts on the three intended steps.
- Docs/workflow only — `git status --porcelain` shows `.github/workflows/ci.yml` and nothing else.

Once this is green and merged, **#252** gets a real green and merges on it — not on an assumption.


MattOrefice:
## This PR carried a second change its body does not describe

The diff is two files, not one:

```
.github/workflows/ci.yml   49 ++++++++--
CLAUDE.md                  11 +++++++
```

The `CLAUDE.md` change is **#252** — the baseline-provenance rule (record the collection total; re-measure after the final commit). It is not described anywhere in the body above.

**Why it is here:** this branch was created from `docs/baseline-provenance-rule` rather than from `main`, so it inherited `9da8855`. Merging this PR made that commit reachable from main and GitHub auto-closed #252 as merged, with no merge commit of its own.

**No verification was skipped.** This PR's CI ran on `1936c84`, which already contained the CLAUDE.md change, and went green in 2m56s. The landed tree `c3b6f0be` is byte-identical to the tree that run tested. But a reader auditing what shipped in this PR would be misled by the body, hence this note.

**The process failure worth recording:** `git checkout -b <new>` inherits whatever HEAD happens to be, and after a previous session's work HEAD was still on the earlier feature branch. The habit that prevents it is `git checkout main && git pull` *before* branching — and the tell that would have caught it here is that the first `git status`/diff on the new branch showed a file the change had no reason to touch.

PR #257, opened 2026-08-18
Derive the crisis-window exclusion instead of naming a sleeve

MattOrefice:
## Root cause

The exclusion set that unlocks the 2008 crisis was written as a literal, `exclude=["US Large Quality"]`. That was correct for the 9-sleeve book, where QUAL (2013-07-18) was the youngest benchmark. The Phase 39 international split (PR #130, `50aa198`, 2026-07-21 14:44 UTC) added **IQLT (2015-01-21)** to the demo book — which starts *later* than QUAL — so dropping QUAL alone stopped buying any history.

The exclusion still ran. It still returned data. It simply stopped working.

`test_extended_history_reaches_2008_live` has failed **every scheduled live-data run since 2026-07-22 — 28 consecutive runs**. Last green was `e9677a8` at 11:21 UTC; the restructure merged at 14:44 UTC; the next run failed. Deselected from the default suite, so no baseline caught it.

Nothing user-facing was ever wrong: `pages/9_Correlations.py:465-468` derives the set at render time and correctly drops both US Large Quality and Intl Quality on the demo book.

## The part worth reading

PR #136 (`20176a1`) **diagnosed this exactly**. Its own commit message says:

> the Extended-history exclusion set (**QUAL alone no longer unlocks 2007 on the demo book**)

It repaired the page, and it edited `tests/test_asset_evaluation.py` in the same commit — adding a test at the file's end while the stale literal sat 80 lines above. One of two consumers repaired, in the same file the other lived in.

## Changes

**1. Derive the exclusion** — `sleeve_first_dates()` and `late_inception_sleeves()` in `src/asset_evaluation.py`, mirroring the page's convention (blends take their earliest leg) so the two derivations cannot disagree about which sleeves are late. The test asserts the *property* the page asserts, not a ticker list.

**2. The vacuous assertion** — `assert default.index.min().year >= 2013` is a lower bound: it passed at 2015 and would pass at 2026. Replaced with a two-sided pin derived from the constraining benchmark's own inception (`c_year <= d_year <= c_year + 1`), plus the **contrast** — the default must NOT reach the crisis and the derived exclusion MUST. Either half alone is satisfiable by a book that makes the Extended-history option pointless.

**3. `src/asset_evaluation.py:138`** — the `"~2007 (DBC inception)"` credit, stale in the same way. SCZ (2007-12-12) is the binding kept leg now. Both occurrences corrected (the second at `average_pairwise_rolling_correlation`).

**4. New offline guard — the structural half, and the actual fix** — `test_page_history_cutoff_matches_module` pins the page's `_RECENT_CUTOFF` to `LATE_INCEPTION_CUTOFF`. Source-text only, no network, no book.

This is the part that matters more than the assertion rewrite. What failed here was not simply a stale literal — it was **a stale literal on the ungated side of a two-consumer split**. The same value lived in two places; the page was covered by render tests in the default suite, the test was `@pytest.mark.live_data` and deselected from it. So when #136 repaired one consumer, the covered one stayed correct and the uncovered one rotted for 28 days, reporting to a badge on a page nobody re-reads.

The new guard runs **where the live test does not** — in the default suite, on every PR. The next time these two derivations drift apart, the failure lands on the gate that blocks a merge rather than on the signal that decays into decoration (#259). Fixing the literal restores today's correctness; moving the check to the gated side is what makes it stay fixed.

**5. `live-data.yml`** — drop the WeasyPrint apt step. Re-verified for *this* job's selection rather than assumed from ci.yml: `weasyprint` is imported function-locally at `src/reports.py:236` and no `live_data`-marked test imports `src.reports`. It never hung here (~5s), but it is the same `apt-get update -qq` that hung twice on one SHA in ci.yml, with `-qq` hiding the reason. `ci.yml`'s comment said "REMOVED" while the identical step was still in the other workflow — it now says which files it describes.

## Verification

The live-data workflow cannot gate this merge, so it was verified on **both books** directly.

| check | result |
|---|---|
| rewritten test, personal 9-sleeve book (local) | passes |
| rewritten test, demo 12-sleeve book (`TRACKER_MODE=demo`) | passes — the configuration failing for 28 runs |
| `tests/test_asset_evaluation.py` offline | 24 passed |
| adjacent files (correlations, render, prose) | 66 passed |

Negative control in demo mode, so the fix is doing the work and not a data shift:

```
derived exclusion set : ['US Large Quality', 'Intl Quality']
OLD exclude=[US Large Quality] -> min 2015-04-17   <=2008? False   ← the CI failure, reproduced
NEW derived exclusion          -> min 2008-03-11   <=2008? True
default (no exclude)           -> min 2015-04-17    >2008? True    ← contrast holds
```

`demo.db`: the demo-mode run drifted it by 235 price and 2 dividend rows. Restored from backup; `fingerprint_db.py` reports IDENTICAL across all 13 tables. For the record, `git diff --stat` called that drift `Bin 28254208 -> 28254208 bytes, 0 insertions(+), 0 deletions(-)`.

A live-data run is dispatched against this branch — see the comment below for its result by SHA.

## Filed, not fixed

- The three-month gap between the stated reach (2007-12-12, the intersection start) and the plotted reach (2008-03-11, after the 60-day burn-in), where the help text implies a pre-crisis baseline that is not in the plot.
- The badge finding: red on the README front page for 28 days with no push signal in any workflow.
- Adjacent and deliberately untouched: `pages/9_Correlations.py:31`'s `_LONG_HISTORY_START` comment also credits DBC, but it justifies a *request* floor rather than naming the binding constraint, and changing it means a render-path change. Noted rather than folded in.


MattOrefice:
## Verification by SHA — both green on `acc20267efd28f8ce8aca344a787056b284af9d9`

The live-data workflow does not gate this PR, so it was dispatched against the branch rather than waited on:

```
gh workflow run live-data.yml --ref fix/derive-the-crisis-window-exclusion
```

| workflow | event | run | headSha | result |
|---|---|---|---|---|
| CI | `pull_request` | [32153465737](https://github.com/MattOrefice/investment-tracker/actions/runs/32153465737) | `acc20267…` | **success** (2m57s) |
| live-data | `workflow_dispatch` | [32153473565](https://github.com/MattOrefice/investment-tracker/actions/runs/32153473565) | `acc20267…` | **success** (1m3s) |

Both record the real branch SHA, so both are re-checkable later. A dispatch run also resolves the workflow file **from the branch**, so run 32153473565 exercised the apt-step removal in the same file it was testing — its step list has no `Install system dependencies (WeasyPrint)`.

**live-data counts:** `36 passed, 1822 deselected` against the failing run's `1 failed, 35 passed, 1821 deselected`. Reconciles exactly: the failing test now passes (35→36 passed) and the new offline guard adds one deselected under the `live_data` filter (1821→1822).

## Filed from this diagnostic

- #258 — the three-month gap between the stated reach (2007-12-12) and the plotted reach (2008-03-11), and the pre-crisis baseline the help text implies but does not show. #246's shape.
- #259 — the badge finding, with four costed options for a push signal.

## One correction to the record

I reported earlier in this session that `gh run list --commit <branch head>` returns nothing for `pull_request` runs because they record a synthetic `refs/pull/N/merge` SHA. **That was wrong.** `--commit` requires the full 40-character SHA and returns empty silently for an abbreviated one:

```
gh run list --commit 1936c84                                    -> empty
gh run list --commit 1936c84d5c964b17bb43e8f49a23a27ca0935029   -> the run
```

So #256's green on `1936c84` was re-checkable all along. The practical warning survives in a more useful form: an empty `--commit` result is *inconclusive*, not negative, and looks exactly like "CI never ran."

Issue #278, opened 2026-08-21
The register's tax axis is the sleeve taxonomy: EQUITY_SLEEVES encodes exposure, not tax character, so a fix keyed on it is wrong by 4x


MattOrefice:
> **Rewritten 2026-08-21 after rendering page 14.** The issue as filed was about a *rate*. It is about an *axis*. The commodity half closed with #210, and what survives is a taxonomy problem the original framing could not see. The original body is preserved at the bottom.

## The axis is wrong, not the rate

`EQUITY_SLEEVES` encodes **exposure**, not **tax character**. `hedged_equity` is in it — correctly, for sizing an account's absorbable-equity capacity, which is what the set exists for — while the income those funds distribute is **ordinary**. So the set that looks like the obvious key for a tax fix is the wrong key, and any fix written against it is wrong in a measurable way.

**Sized against the rendered KPI.** Page 14 renders `Annual tax drag (A/B/D)` = **$276** (rebuild reproduces $275.96):

| fix | result | error |
|---|---|---|
| tax every `EQUITY_SLEEVES` row at LTCG | $56.21 moved (15.5%) | **wrong by 4×** |
| tax by actual character | **$7.34 moved (2.7%)**, KPI $276 → $268.62 | — |

The gap is `hedged_equity`, which is **39% of the whole KPI**:

| symbol | case | drag | why `ordinary` is right |
|---|---|---|---|
| JEPQ | B | $41.95 | ELN / covered-call income is ordinary |
| JEPI | B | $40.42 | same |
| JHEQX | B | $25.25 | index options, 60/40 — not qualified |
| | | **$107.62** | |

**The repo already says this, in prose the page renders**: *"5 ordinary-income holdings worth \$20,133 sit in the TOD account — 4 of them — a hedged-equity mutual fund, a liquid-alternatives fund, a global allocation fund, and a core-plus bond ETF — throw ordinary income into your highest-taxed wrapper."* And the Roth cleanup card: *"their heavy ordinary income is sheltered for free there."* The distinction exists, is authored, and is rendered to the reader. It is the *register* that cannot represent it.

## What is actually mis-rated, measured

| treatment | status | moves |
|---|---|---|
| **Commodities** | **closed by #210.** `real_assets_commodities` is in `NOT_MODELLED_SLEEVES`; PDBC's `annual_benefit` is `None`, so **no rate is applied to it at all**. This issue's original headline is spent. | $0 |
[...]

Issue #249, opened 2026-08-18
The Bitcoin case study puts seven independent sub-analyses behind one banner


MattOrefice:
**Coarse disclosure over fine-grained failures** — worth naming as its own shape, because it is not a missing disclosure and not a fabricated value.

`_build_asset_eval_section` (`src/reports.py:1216`) runs seven independent computations, each in its own `try/except: pass`:

```
:1264  ae.build_univariate_table()
:1309  ae.compute_full_sample_correlations(...)
:1352  ae.compute_rolling_correlation(...)
:1383  ae.compute_mv_analysis(...)
:1409  ae.compute_marginal_sharpe_curve(...)
:1427  ae.compute_drawdown_sensitivity(...)
:1473  the for/against argument assembly
```

The section has exactly **one** disclosure, at `templates/quarterly_report.html:533`:

> Asset evaluation data unavailable — requires market data access.

That fires when the section as a whole has nothing. It cannot express **partial** failure. If six sub-analyses succeed and the drawdown sensitivity raises, the PDF renders six results and says nothing about the seventh — the reader sees a complete-looking section that is missing a piece, with no way to know which piece or that any is missing at all.

## Why it is a distinct shape

The other defects in this workstream were: a fabricated value (`50.0`), a false label (a window that was not used), or a swallowed reason (one sentence for two states). This is none of those. Every rendered number is real and correctly labelled. The defect is **granularity**: the disclosure's resolution is coarser than the failure's.

The general form, worth watching for elsewhere: **a section-level banner cannot disclose a sub-section failure**, and adding sub-analyses to a section without subdividing its disclosure silently widens the gap. Seven is already enough that "the section is fine" and "the section is complete" have come apart.

## Not urgent

The Bitcoin case study is a reference exhibit rather than a decision input, so the does-a-reader-act test is weak here — which is why this is filed rather than folded into the failure-paths PR. The fix is mechanical whenever it is done (a per-analysis availability flag, or a list of names that failed, rendered as the existing `price_gaps` pattern does).

Issue #228, opened 2026-08-17
The Roth deploy's developed-international exclusion is undisclosed in rendered prose (unblocked: #221 decided 2026-08-17)


MattOrefice:
**Blocked on #221.** The disclosure cannot be written until the rule is settled, because right now it would document a gap rather than a decision.

## The gap

The Roth deploy card excludes broad and developed international from its target set. The reason exists — a Roth forfeits the foreign tax credit — but it lives **only** in places a reader never sees:

- `src/location_config.py:285-286` — a code comment
- `src/location_config.py:317-322` — a docstring on the taxable priority map, unrendered

Nothing in the page's rendered prose says why `intl_developed` and `intl_all_exus` are absent from a deploy table. Meanwhile an international sleeve **is** in that table:

```
ticker           sleeve  dollar
  SPHQ us_large_quality 3032.11
  AVUV   us_small_value 1927.52
   VTV   us_large_value 1283.18
  IEMG emerging_markets 1071.77
```

So a reader who happens to know the FTC argument sees the table contradict it, and a reader who does not is never told the argument exists. The page is silent in both directions.

## Why it is blocked rather than actionable

Whatever the disclosure says depends entirely on #221's answer:

- **If EM is excluded** (implementing `b6b8774`'s stated intent), the prose is simple: international sits in taxable because only a taxable holder can credit the foreign tax withheld, so the Roth deploy skips it. Clean, and it matches the table.
- **If EM is kept**, the prose has to explain why one international sleeve is an exception to a rule stated in terms of the FTC — and that rationale does not exist yet. EM carries *higher* foreign withholding than developed markets, so the argument cannot be the credit; it would have to be growth outweighing the forfeited credit, which is a judgement nobody has made.

Writing the disclosure now would mean documenting the current inconsistency as though it were the design. That is the failure mode #191 closed on the assumed-yield table, arriving from the other direction.

## Where it would go

The Asset Location page's Assumptions expander already discloses two other modelling choices in exactly this form — that scores are authored, and that sleeve deploy targets are an ordinal ranking rather than a return forecast. A third clause naming the FTC exclusion belongs there, in one sentence, alongside them.

## Scope note

Prose only, once unblocked. No behaviour change — the exclusion mechanism (absence from `_ROTH_PRIORITY`) is not in question here; only whether the page says so.

MattOrefice:
## Unblocked — #221 is settled

[...]

Issue #187, opened 2026-08-13
12 render tests are demo-mode-only: the suite cannot be run in personal mode


MattOrefice:
## The gap

A plain `python -m pytest -q` on this machine is **13 failed**. Twelve of those are render tests that pass in demo mode (CI is green) and fail in personal mode:

| module | count | tests |
|---|---|---|
| `tests/render/test_performance_render.py` | 5 | `test_period_returns_no_suppression_on_full_history`, `test_two_stage_reconciliation_uses_price_series`, `test_two_stage_attribution_section_renders`, `test_stage_reconciliation_gap_caption_appears_under_gap`, `test_naive_gap_notice_and_unavailable_tiles_under_total_gap` |
| `tests/render/test_factor_profile_render.py` | 4 | `test_global_factors_discontinued_in_methodology`, `test_morningstar_independence_statement_present`, `test_fi_sleeve_carhart_note_present`, `test_regression_window_lag_caption_present` |
| `tests/render/test_trade_log_render.py` | 3 | `test_themes_tab_framing_lead`, `test_themes_tab_reits_example`, `test_core_fi_thesis_lead` |

The 13th is #177 (`test_exactly_the_saa_tickers_are_flagged`), which is deliberate and load-bearing.

## What it is not

Ruled out while landing #185:

- **Not a regression.** All 12 fail identically at `fd82346` with the branch stashed.
- **Not DB drift.** They fail unchanged after restoring both `tracker.db` and `demo.db` from pre-run snapshots (`sha256` verified identical).
- **Not degraded personal data.** `tracker.db` has 14/15 `asset_classes.rationale` populated, and the specific pinned string these tests look for ("Duration as recession ballast…") is present in **both** DBs. The content is there; the personal-mode page render doesn't surface it.
- **Not an env-var accident.** Same 12 fail with `TRACKER_MODE` unset (`.env` pins personal) and with `TRACKER_MODE=personal` explicit.

## Why it matters

Several of these tests carry docstrings describing themselves as DB-dependent regression anchors on authored content (`asset_classes.rationale`, factor methodology captions, two-stage attribution). In demo mode they anchor demo data. In the mode the owner actually runs the app in, they don't run at all — they fail, which is worse than skipping, because it buries the one deliberate red (#177) in noise and makes every local suite run require triage before it can be read.

[...]

PR #281, opened 2026-08-21
Name whose drawdown the case against is talking about

### LINKED ISSUE #276: The decision framework's two sides are asymmetric in kind: args_for is derived, three of four args_against are hardcoded
Split out of #249 rather than folded in, because disclosing the asymmetry is not the same as removing it — and removing it is a different kind of change.

## The structure

In `_build_asset_eval_section`'s decision-framework block:

| side | entries | depends on |
|---|---|---|
| `args_for` | **1**, conditional | the marginal-Sharpe curve (5g), plus an MVO cross-check from 5f |
| `args_against` | **1** conditional | the full-sample correlation (5b) |
| `args_against` | **3** unconditional | *nothing* — `args_against.extend([...])` with three hardcoded strings |

So one side is entirely derived and the other is three-quarters asserted. **Only the derived side can fail**, which makes any failure directional: the case *for* can be silenced by a computation error while the case *against* renders in full, every time.

#249 fixed the reader-facing half of that — the render no longer says "None identified from available data" when the curve did not run; it says the case was **not assessed** and states that the arguments against "were not weighed against anything". That is a disclosure, and the structure it discloses is still there.

## The question, which I do not think #249 should have decided

**Should the two sides be symmetric in kind?** Three readings, and they lead to different work:

1. **They are already correct as they are.** The three hardcoded arguments (an >80% historical drawdown, commodity tax treatment, no intrinsic cash-flow anchor) are *editorial* claims about the asset class, not measurements. They do not depend on this portfolio's data and would be true in any report. Under this reading the asymmetry is honest — one side is evidence, the other is a standing argument — and the fix is to *label* them as such rather than make them conditional.
2. **They should be derived too.** The drawdown claim in particular is measurable — 5h computes a drawdown sweep — so asserting ">80%" as a constant while a live computation of the same quantity sits three blocks above is the [[same-artifact-one-half-discloses]] shape #245 was about. Under this reading they should be computed and made conditional.
3. **`args_for` should have a standing side too.** The asymmetry could be closed from the other end — a symmetric editorial argument *for* inclusion — which would remove the directionality without making anything conditional.

**(1) and (2) disagree about whether the three strings are claims about Bitcoin or claims about this portfolio's Bitcoin exposure**, and that is a product question, not a code one.

## Why it is not urgent

Nothing renders wrong today: after #249 the failed-curve case says so explicitly. This is about whether a reader can tell that the two sides of the comparison were produced by different means — which the report currently does not say anywhere.

## Pinned meanwhile

`tests/test_asset_eval_subanalysis_disclosure.py::test_args_against_still_survives_every_failure_and_that_is_filed` asserts the unconditional count is exactly 3, so the remainder cannot drift while this is open. If the count changes deliberately, that test names this issue.

## Related

- #249 — the disclosure half, which this was split out of
- #245 — one artifact where one half discloses and the other fabricates; same family

MattOrefice:
Closes #276.

## The number was true. That was never the problem.

`"Maximum historical drawdown exceeding 80%"` was hardcoded, and **correct** — BTC's own max drawdown is **−81.5%** over the report's own sample (2018-01-02 → 2026-08-21, 3,153 obs), measurable from a `btc_ret` already in scope three blocks above it. So this could not be fixed by deriving the figure and stopping: that would have replaced a right number with the same right number and left the thing that actually misleads.

**What misleads is adjacency.** The template renders `<h3>Drawdown Sensitivity</h3>` and then, as the very next `<h3>`, `Decision Framework`:

| BTC alloc | CAGR | Max DD | Sharpe | 2022 MDD |
|---|---|---|---|---|
| 0.0% | 10.1% | **−31.3%** | 0.35 | −23.3% |
| 10.0% | 12.1% | **−29.8%** | 0.45 | **−26.1%** |

> • Maximum historical drawdown exceeding 80% …

That column is the **portfolio's**, and it *improves* as BTC rises. The bullet is **Bitcoin's**, four times larger. Neither said whose. A reader sees drawdown falling, then a bullet saying it exceeds 80%.

## What the bullet says now

> Bitcoin's own maximum drawdown is **−81.5%** over 2018-01-01–present — **not the −31.3% in the table above, which is the portfolio's**. Its 2022 drawdown of −66.9% came while equities and bonds fell together, deepening the portfolio's own 2022 drawdown from −23.3% at 0% allocation to −26.1% at 10.0% — no diversification benefit when it was most needed

The old second clause asserted qualitatively that 2022 offered no diversification benefit. The sweep **measured** exactly that, one table up. It now says by how much.

## Two forms, because the halves fail differently

`btc_ret` is guaranteed non-empty past the loader guard (`reports.py` returns `no_data` before 5j otherwise), so Bitcoin's figure always renders. The portfolio contrast comes from 5h, which has its own `except`. Reduced form:

> Bitcoin's own maximum drawdown is −81.5% over 2018-01-01–present. The portfolio-level comparison — how much a BTC allocation deepened the 2022 drawdown — is not available this run: the drawdown sweep did not produce data. **Absent, not zero**

**"Absent, not zero" is load-bearing.** A contrast that quietly vanishes reads as *no* contrast — as though the allocation had cost nothing, which is the opposite of what the full form reports.

## The cheapest fix here was found while writing the expensive one

`Max DD` → **`Portfolio Max DD`**, `2022 MDD` → **`Portfolio 2022 MDD`**. Two `<th>` cells. Naming the subject at the point of reading would have made the adjacency legible **with no prose change at all** — the whole derivation above is the expensive way to say what two column headers say for free. Worth recording because the search that found it started from the sentence, not the table.

## The caption called realized history a scenario

It read *"2022 stress scenario: equities −20%, bonds −15%, BTC −65% simultaneously"* beside a column computed from **actual 2022 returns** (`stress = combined.loc["2022-01-01":"2022-12-31"]`). Calling a measured column a scenario invites reading it as hypothetical — the **inverse** of the `>80%` defect, and arguably worse: one is an unlabelled measurement, the other a measurement mislabelled as a projection.

It now says realized history and templates Bitcoin's measured **−66.9%**. The equity and bond figures are **dropped, not derived**: two of the three were never verified, and none of them is what the column computes — so deriving accurate sleeve-level numbers would replace unverified figures with correct ones that still describe something else.

## Three entries, three kinds, three dispositions

The issue assumed one disposition for all three. The render says otherwise.

| entry | kind | done |
|---|---|---|
| drawdown | a **measurement** written as prose | **derived**, subject and window named |
| commodity tax treatment | a **statement of tax law** | **frozen**, comment naming #278 |
| no cash-flow anchor | an **editorial judgement** | **labelled** "Standing argument, not a measurement" |

**#2 is deliberately untouched.** It is the third in-repo copy of the commodity tax-character claim, and #278 asks whether the asset-location register should grow a rate vocabulary — if it does, this string must *derive* from it. Editing it here first mints a fourth copy, which is exactly how this family reproduces (four copies of one FTC sentence turned up on #228 the same day).

## The pinned test asked to be updated, and it was

`test_args_against_still_survives_every_failure_and_that_is_filed` asserted the hardcoded count was exactly 3 and said *"if that was deliberate, update #276."* It was: the drawdown left the `extend([...])` block, so the count is **2**. A second test now pins that the drawdown is appended **unconditionally and not inside an `if`** — deriving it must not make the case against conditional, which would shrink the arguments against on exactly the runs a reader can least check.

**Recorded: reading that count as "three of four are hardcoded" was always brittle.** The fourth entry is the correlation argument, gated on `> 0.3`, which measured **0.31** on 2026-08-21. One hundredth lower and the ratio reads "three of three" — a *worse* asymmetry described by a *better*-sounding number, with nothing changed in the code.

## Mutation proof — 10 mutants, all killed by a named victim

Aimed at the **rendered sentence**, not the computation: the computation was already correct, so an arithmetic mutant proves nothing about this PR.

| | mutant | killed by |
|---|---|---|
| M1 | restore the hardcoded ">80%" head | `..._names_bitcoin_as_the_subject...`, `..._states_the_window`, `..._tracks_the_data_rather_than_a_literal` |
| M2 | drop the clause naming the table's figure as the portfolio's | `..._names_the_TABLE_figure_as_the_portfolios` |
| M3 | drop "Absent, not zero" | `..._declares_the_contrast_absent_not_zero` |
| M4 | append the drawdown only when the sweep ran | `..._did_not_make_the_case_against_conditional` +3 |
| M5 | unlabel the standing argument | `..._is_labelled_as_standing` |
| M6 | delete the comment naming #278 | `..._is_frozen_and_says_why` |
| M7 | make `_max_drawdown` return a magnitude | `..._is_signed` |
| M8 | call the measured column a scenario again | `..._does_not_call_the_measured_column_a_scenario` |
| M9 | restore the unverified sleeve figures | `..._drops_the_unverified_sleeve_figures` +2 |
| M10 | revert the column headers | `..._name_the_portfolio_as_the_subject` |

The harness asserts each edit actually changed the file, counts `ERROR` as a kill alongside `FAILED`, and checks the kill came from the named victim rather than collateral.

## Presence is not use

Every direct test of `_drawdown_argument` would pass on a builder that never called it — the same shape as a source pin satisfied by a helper's `def`. Two tests run the real builder with a **synthetic sweep** whose figures cannot coincide with the real book (−44.0% / −11.0% / −19.0%), so their appearance in the bullet proves `_port` was populated from the sweep *and* handed to the helper.

## A second guard earned its keep

`test_the_harness_context_leaves_nothing_undefined_inside_the_section` exists in two files and went red on the full suite the moment the builder grew `btc_2022_mdd` — which is what it is for. #249 added the same note beside `unavailable` for the same reason. Second catch; recorded in the file, because a premise assertion that looks like bookkeeping is the kind that gets deleted.

## Verification

- **Collection:** 1994 selected (2034 − 40 deselected)
- **Guards UP**, personal mode, plain `pytest` on the final tree `7a8ac16`: **13 failed / 1954 passed / 27 skipped** — 13 + 1954 + 27 = **1994**, reconciles
- The 13 are #187's twelve mode-sensitive render tests plus #177, both deliberate and load-bearing
- `demo.db` fingerprint identical across all 13 tables; `tracker.db` sha256 unchanged


MattOrefice:
**Guards-DOWN convergence arm**, same tree `7a8ac16`: **13 failed / 1954 passed / 27 skipped**, 1994 collected.

Issue #250, opened 2026-08-18
_build_asset_eval_section's two _empty returns are unrendered — verify before assuming they behave like the attribution one


MattOrefice:
`src/reports.py:1243` and `:1246` return an `_empty` dict from `_build_asset_eval_section`, structurally similar to the attribution section's `_empty` that the failure-paths PR just gave a disposition field.

**Do not assume they behave the same.** The attribution one was analysed from source as "the section vanishes and fabricated `+0.0` totals render", and an in-memory render showed **both halves of that wrong**: the outer template gate has an else branch, and the totals never reach the page because the table body is gated out. Source analysis produced a confident wrong picture twice in that investigation; the render is what settled it.

So the work here is: render the PDF in-memory with each of these two paths forced, and read what appears. Then decide whether a disposition field is warranted, using the same test — does one sentence cover two materially different states?

Method that worked, for reuse: patch `reports._render_pdf` to capture the HTML string and return a stub, patch the computation to raise, call `generate_quarterly_report_bytes`. WeasyPrint is never invoked and nothing is written to `data/reports/`.

Related: the seven-sub-analyses-behind-one-banner issue covers the same section from the granularity angle.

Issue #230, opened 2026-08-17
The Roth intl exclusion's off-SAA argument expires when the 9→12 split lands


MattOrefice:
Split out of #229 (which settled #221). **Dated, not urgent** â€” it becomes live only when the international restructure completes.

## The tension

After #229, `location_config.py`'s Roth rationale distinguishes two populations:

- **excluded** â€” `intl_developed`, `intl_all_exus`, and the three tilt sleeves: off-SAA target-$0 rows pre-migration, so a Roth buy moves away from the SAA; plus the phase-46 gate and Roth scarcity
- **kept** â€” `emerging_markets`: a targeted SAA sleeve at 9.18%, so none of those reach it

And `location_config.py:317-322` says the tilts are kept out of a *shelter* because the forgone FTC is a real cost. A Roth is also a shelter, so read literally that reasoning would exclude EM too.

**Today the two reconcile on the off-SAA distinction.** EM is targeted; the tilts are not.

## Why it expires

When the 9â†’12 split lands, the tilts become targeted SAA sleeves â€” the demo book already carries them at **Intl Quality 6.25% / Intl Large Value 3.75% / Intl Small Value 3.33%**. At that point they stop differing from EM on the off-SAA axis, and the exclusion has to rest on one of:

1. **Roth scarcity alone** â€” defensible, but it is a weaker argument than the current three-part one, and it would apply to EM equally unless growth-premium ranking is the tiebreak (which is what `_ROTH_PRIORITY` already is).
2. **A revisit** â€” decide whether one or more tilt sleeves belongs in the Roth map at all now that they are targeted, on the same growth-versus-forgone-credit trade EM was decided on.

Either way the current rationale stops being accurate, which is why the comment writes it as an explicit DATED clause rather than a timeless one.

## Not fixed now, deliberately

The tilts are unheld and the split is held pending the $227 carrier buys in `acct_01`, so nothing renders differently today. Writing the post-split rationale now would mean authoring a rule for a taxonomy that does not exist yet â€” and if the answer is option 2, it is a portfolio decision rather than a comment.

## Related

- #229 â€” the rewrite that surfaced this and carries the dated clause
- #221 â€” the decision to keep EM (closed by #229)
- #228 â€” the rendered disclosure, unblocked by #229; note it would need the same revisit, since a disclosure written pre-split states a rule that changes
[...]

Issue #221, opened 2026-08-17
emerging_markets is in the Roth deploy map against b6b8774's stated intent — semantics question, not a comment defect


MattOrefice:
**Rewritten. This is not a comment defect — it is an incomplete edit, and correcting the comment would encode a bug as intent.**

Originally filed as *"a prose defect whose source is a code comment"*: `location_config.py:285-286` says *"International is absent: a Roth forfeits the foreign tax credit"* while `emerging_markets` sits at rank 2 in the same map. Reading the commit rather than the comment reverses that framing.

---

## The diff, which settles it

`b6b8774` ("Per-account-type sleeve priority; delete deploy alias + exclusion set") introduced `_ROTH_PRIORITY`:

```diff
-    "emerging_markets":         2,        # the OLD flat SLEEVE_LOCATION_PRIORITY
-    "intl_developed":           6,
-    "intl_all_exus":            6,
+_ROTH_PRIORITY: dict[str, int] = {
+    "emerging_markets": 2,               # carried over UNCHANGED into the Roth map
+    "intl_all_exus":    1,               # these two moved to the TAXABLE map
+    "intl_developed":   1,
```

It **removed** `intl_developed` and `intl_all_exus` from the Roth map and **left `emerging_markets` behind.** And it states the intent twice in its own message:

> roth_ira/hsa rank by expected return (us_small_value 1, **intl absent: a Roth forfeits the FTC**)

> Eligibility is structural (only roth-map sleeves with an SAA ticker appear), so cash/FI/real-assets/hedged/**international** can never surface.

So the intended rule is *"no international in the Roth."* The implementation lets emerging markets through. **`emerging_markets: 2` is the defect**, not the sentence describing it.

## The rendered consequence

IEMG is in the Roth deploy table **today**, out of \$7,314.58 of idle cash:

```
ticker           sleeve  dollar
  SPHQ us_large_quality 3032.11
  AVUV   us_small_value 1927.52
   VTV   us_large_value 1283.18
  IEMG emerging_markets 1071.77   <- international, in the Roth
```

\$1,071.77 forfeiting exactly the foreign tax credit the exclusion exists to protect.

## Both answers change something

**Exclude EM** (implement the stated intent) — IEMG leaves the table and \$1,071.77 redistributes across the remaining three sleeves. A behaviour change to a rendered trade recommendation.

[...]

Issue #177, opened 2026-08-13
securities.is_in_saa is unguarded: carriers reached the personal book through load_household_securities, which no taxonomy guard watches


MattOrefice:
`securities.is_in_saa` has no guard. The personal book's 9-sleeve taxonomy is protected three ways; the SAA **flags** are protected zero ways, and they have already diverged.

## The gap

The CHANGELOG entry *"Deferred: a from-scratch personal reseed would jump the taxonomy to 12"* (2026-07-21) records the protection accurately:

> `seed_saa.seed()` skips a non-empty `asset_classes`, and the Phase 39 migrations are demo-only (the taxonomy migration exposes only `main()`, which the bootstrap skips and which hardcodes demo.db; the restructure migration guards on `db_path.name == "demo.db"`). So the real household book stays 9 sleeves, **guarded three ways**.

All three guard `asset_classes`. None watches `securities`.

`cf931cf` ("Update the suite for the twelve-sleeve taxonomy", 2026-07-19) added IDHQ/AVIV/AVDV to `data/seed/securities_household.csv` with `is_in_saa=true`. That file is consumed by `load_household_securities` (`src/seed/securities_loader.py:37`), whose own docstring reads **"Personal-mode only: run against tracker.db. Do not run against demo.db"**, called from `bootstrap_personal_db()` at `src/bootstrap.py:109`. So the carriers entered the **personal** book through a personal-only loader that none of the three guards covers.

`5e009db` reverted `SAA_TICKERS` in `tests/test_securities_mapping.py` back to 11 the following day — deliberately, per its commit message ("the personal tracker keeps its coherent 9-sleeve taxonomy until real trades fund the tilts") — but did not revert the seed row.

## Current state (measured 2026-08-12)

`tracker.db` has **14** tickers at `is_in_saa=1`: the documented 11 plus IDHQ, AVIV, AVDV. All three carry **`asset_class_id=15` = 'Other / Non-SAA'** and have **0 trades**. They are flagged as SAA carriers in a book whose taxonomy contains no such sleeves.

The CHANGELOG's own warning — *"so taxonomy and holdings never diverge on a clean rebuild"* — describes a hypothetical fresh reseed. The divergence is already present on the **existing** book.

## Deliberately not fixed

[...]

PR #282, opened 2026-08-22
Say why the Roth deploy skips developed international

### LINKED ISSUE #228: The Roth deploy's developed-international exclusion is undisclosed in rendered prose (unblocked: #221 decided 2026-08-17)

MattOrefice:
Closes #228. Prose only — no behaviour change, asserted rather than claimed.

#221 settled the rule on 2026-08-17 (`da2e590`: emerging markets **stays**, the code was right and the rationale was wrong) and **nothing rendered said so**. The Roth deploy buys IEMG while `intl_developed` and `intl_all_exus` are absent from the map, so a reader who knew the foreign-tax-credit argument saw the table contradict it, and a reader who did not was never told the argument existed. Silent in both directions.

## 1. It opens by ruling the FTC out, rather than omitting it

> **Why international is split across wrappers.** The Roth deploy buys emerging markets but not developed or broad international, and **the reason is *not* the foreign tax credit** — a Roth forfeits that on emerging markets too, **and by more**, since emerging-market withholding runs higher than developed's. The developed sleeves are excluded because, today, they sit off the SAA as target-\$0 rows, so buying them in the Roth is movement *away* from the target on this page's own terms; because the gate that would bring them onto the SAA counts carrier positions in the taxable book only, so a Roth trade cannot open it; and because Roth space is the scarcest shelter with the 401(k) rollover still inbound. **The first of those three expires when the international split lands** — after which the exclusion rests on scarcity alone, or is revisited. Emerging markets is a targeted SAA sleeve and currently underweight, so none of the three reaches it: the credit it forfeits is a real cost, accepted rather than avoided.

**Two failure modes, not one, and the clause has to dodge both.**

The first is the one #228's own body walked into: it drafted the disclosure as *"international sits in taxable because only a taxable holder can credit the foreign tax withheld, so the Roth deploy skips it"* — verbatim the blanket framing `da2e590` deleted from three comments for being "too coarse in both directions." Writing it would have restored, **on the one surface a reader actually sees**, the claim just removed from three they don't.

The second is subtler: **omitting the FTC is not enough either.** The target reader is precisely the one who already knows the argument and is looking at IEMG sitting in the deploy table. Silence leaves them holding the contradiction with nothing to resolve it.

So the clause states the rebuttal **and the fact that makes it decisive**: emerging markets forfeits *more* credit than the developed sleeves and is kept anyway — so the credit cannot be what separates them. Pinned by `test_the_clause_rules_the_ftc_out_rather_than_omitting_it` and `test_the_rebuttal_gives_the_fact_that_makes_it_decisive`, and killed by M1 (omit) and M2 (assert without the reason) separately, because those are two different regressions.

## 2. The expiry is stated, not implied

One of the three reasons — that the developed sleeves render off-SAA as target-\$0 rows — **stops being true when the 9→12 split lands**. Stated, #230 arrives as a scheduled revision. Implied, the sentence goes quietly false and nothing goes red.

`test_the_clause_states_its_own_expiry` pins it; `test_the_clause_gives_all_three_reasons_not_just_the_durable_ones` blocks the tempting shortcut of dropping the expiring reason — a clause that gave only the two durable ones would need no expiry note **and would be wrong today**, because the exclusion does not rest on scarcity yet.

## 3. The fourth copy points rather than restates

`da2e590`'s own message said *"three places"*. There were four — the docstring of the function that builds the deploy table, written 2026-07-11 at `e3d0d24`, predating both the comment it echoed and the correction.

**Restating is the mechanism that produced four copies of the wrong claim, and a fifth restatement of the right one is the same behaviour with better luck.** So the correction redirects to the single record above `_ROTH_PRIORITY` rather than reproducing the three reasons. `test_the_corrected_docstring_points_rather_than_restates` asserts it points *and* that it has not quietly become a fifth copy — it fails if `phase-46`, `76,147` or `9.18%` appear in that docstring.

The retired wording is **described, not quoted**: quoting it would put the exact string back into a file the sweep scans, so the sweep would flag its own correction — the checker inside its own sample.

**The evidence:** a post-fix sweep for the blanket claim returns **nothing outside `location_config.py`**, where the one correct statement lives. That sweep is now an assertion (`test_no_blanket_ftc_claim_survives_outside_the_config`) rather than a one-off grep.

## Direction without numbers, and the reasoning recorded beside it

The clause says emerging-market withholding *"runs higher than developed's"* and omits the ~20–25% vs ~15% behind it. Those are **authored country-mix estimates**, not measurements from anything in this repo.

**The direction survives them being loose** — EM withholds more than developed under any plausible version of either number, which is all the argument needs. **The figures do not.** Printed on a page whose own Assumptions promise live-or-refuse for every dollar shown, they would read as measured, and a reader could not tell them from the computed figures beside them.

That reasoning is in a source comment beside the sentence, ending with the question to answer first: *what measured them*, not whether they are roughly right. `test_the_reasoning_for_withholding_direction_is_findable_in_source` keeps it there — M5 deletes the note and dies.

No dollar amount is hardcoded in the clause at all, for the same reason. `test_the_clause_hardcodes_no_dollar_figure` allows `\$0` (a *category* of row) and blocks any amount with a leading non-zero digit.

## Mutation proof — 11 mutants, all killed by their named victim

**M7 SURVIVED on the first pass, and the instrument was at fault rather than the tests.**

The sweep searched raw file text for a contiguous phrase. A reintroduction **wrapped across two lines** — which is exactly how it would come back, in a docstring or a comment — was invisible to it. The test was measuring line wrapping, not content. Same trap as the #250 mutation harness and the `real_assets_commodities` evidence check, both of which hit it before this did.

Fixed by collapsing whitespace and stripping comment prefixes before searching, and the single mutant became three, because *"the claim came back"* has three realistic spellings:

| | mutant | |
|---|---|---|
| M1 | omit the FTC instead of rebutting it | KILLED |
| M2 | assert "not the FTC" without the decisive fact | KILLED |
| M3 | drop the stated expiry | KILLED |
| M4 | print the authored withholding figures | KILLED |
| M5 | delete the direction-without-numbers reasoning | KILLED |
| M6 | hardcode the rollover balance | KILLED |
| **M7** | reintroduce the claim **wrapped across two lines** | KILLED *(survived before the fix)* |
| **M7b** | reintroduce it on one line | KILLED |
| **M7c** | reintroduce it in a wrapped comment | KILLED |
| M8 | make the docstring restate instead of point | KILLED |
| M9 | drop the reason the FTC cannot be the rule | KILLED |

A second instrument note: `_prose()` reads the clause through `ast`, not by grepping the page. Its first version did grep, and **all seven prose assertions failed at once** — the clause is ~20 adjacent string literals, so every reader-facing phrase is cut by `"` seams, and `target-\\$0` on disk is `target-\$0` to a reader. Grepping would have tested the line wrapping. The comment block above the clause also quotes the retired wording, so a raw search would find the test's subject inside its own documentation.

## Verification

- **Collection:** 2006 selected (2046 − 40 deselected)
- **Guards UP**, personal mode, plain `pytest` at `c519038`: **13 failed / 1966 passed / 27 skipped** — 13 + 1966 + 27 = **2006**, reconciles
- The 13 are #187's twelve mode-sensitive render tests plus #177, both deliberate
- Rendered in local Streamlit (personal mode) before the wording was settled; `demo.db` fingerprint identical across all 13 tables, `tracker.db` sha256 unchanged

## Scope

It did not grow — the risk flagged when this was picked up, given it sat mislabelled for four days. Two files plus the test: one Assumptions clause, one docstring correction. `test_the_eligibility_rule_itself_is_unchanged` asserts the structural rule is untouched (`emerging_markets` in the Roth map; the five international sleeves out) rather than the PR body claiming it.


MattOrefice:
**Guards-DOWN convergence arm**, same tree `c519038`: **13 failed / 1966 passed / 27 skipped**, 2006 selected.

Issue #284, opened 2026-08-22
drip.py and reports.py contradict each other on PDBC's tax character, and only the rendered one is used to argue a case


MattOrefice:
Found while scoping #278's design. Two in-repo statements of PDBC's tax character, neither derived from anything, **and they contradict each other.**

## The two statements

**`src/drip.py:8-11`:**

> PDBC NOTE: PDBC distributions have mixed tax character (**qualified dividend**, return of capital, ordinary income). Phase 17 treats PDBC identically to other ETFs — distributions are reinvested at payment-date adj_close. Tax-character splitting for PDBC cost basis is deferred to a future phase.

**`src/reports.py:1694-1695`** — rendered in the quarterly PDF's Decision Framework, as a standing argument against commodity inclusion:

> Commodity tax treatment: short-term ordinary income / long-term capital gains, **no qualified-dividend treatment** — unfavorable vs. equity ETFs in taxable accounts

One says PDBC's distributions include qualified dividends. The other says commodity treatment has **no** qualified-dividend treatment and makes that the argument. They cannot both be right about the same fund.

## Why it matters beyond being untidy

**Only one of them is rendered.** `reports.py`'s version reaches a reader, in a PDF, as a reason not to hold the asset class. `drip.py`'s version is a module docstring. So the version a decision gets made on is the one nobody can check against the other, and the disagreement is invisible from the surface where it counts.

**Neither is derived.** Both are authored prose. Nothing in the repo computes the character split, and `drip.py` says so explicitly — *"Tax-character splitting for PDBC cost basis is deferred to a future phase."* So there is no third artifact to break the tie.

**#276 counted one and missed the other.** That issue identified `reports.py:1694` as the third in-repo statement of commodity tax character and froze it pending #278's vocabulary decision. It did not notice `drip.py:8` disagrees with it. The count was wrong and, more importantly, the *character* of the finding was wrong: these are not redundant copies of one claim, they are inconsistent copies.

## This is #278's B2 argument, made by the codebase rather than by me

#278 asks whether tax character should be a declared vocabulary or a handful of exceptions. The case against exceptions was that scattered prose statements drift apart and nothing reconciles them.

**That already happened here, before a single exception was added.** Two files, two authors' intentions, one fund, no shared declaration, and the statements diverged with nothing to catch it. A vocabulary would make this contradiction unrepresentable — there would be one declared character for the sleeve and both surfaces would read it.

[...]

Issue #283, opened 2026-08-22
Page 8 renders a stale DC jurisdiction on VGIT's Treasury-exemption rationale — a DATA fix (seeder is insert-only; text lives in tracked demo.db and deploys publicly)


MattOrefice:
Found while scoping #278's design. **Live today, and independent of whatever #278 decides** — the register could grow a full tax-character vocabulary or none at all and this contradiction would survive either.

## Two rendered pages of the same app disagree about one security

**Page 8 (Research)** renders VGIT's holding rationale. Verbatim from a personal-mode render, 2026-08-22:

> VGIT delivers intermediate-term US Treasury exposure at 0.04% versus IEF's 0.15%. […] **Critically, Treasury interest income is exempt from state and local taxes — a real after-tax advantage in a high-income-tax jurisdiction like DC over investment-grade corporate bond funds with similar yield.** Cost minimization is the dominant selection criterion […]

Source: `src/seed_securities.py:237-239`, rendered through `pages/8_Research.py:287`.

**Page 14 (Asset Location)** charges VGIT's income the full combined rate:

```python
income_rate = state_only if _is_federally_exempt(sleeve) else ordinary   # household.py:888
```

`core_fi_treasury` is not in `FEDERALLY_EXEMPT_SLEEVES`, so VGIT gets `ordinary` = `federal_marginal + state_marginal` = **25.07%**, state included. The correct rate on Treasury interest is federal only, **22.00%**.

## Why this is the cross-surface shape at its sharpest

The usual form of this defect is prose asserting a fact that some computation elsewhere ignores. Here it is worse in two specific ways:

1. **The ignored fact is the stated reason the security was chosen.** The rationale does not mention the exemption in passing — it says *"Critically"*, and names it as a selection criterion against corporate bonds of similar yield. The register then prices VGIT exactly as if it were one of those corporate bonds.
2. **Both statements are rendered.** This is not a comment disagreeing with code. A reader can open page 8, read that the state exemption is a real after-tax advantage, open page 14, and see the drag computed with the state tax applied. Nothing on either page acknowledges the other.

## Size

**$0.30/yr**, on the $276 drag KPI. Deliberately quoted as the *wrong* measure of it: VGIT and SCHP total $227 of positions today, so the figure is small because the book is, not because the rule is. The same rule misprices a Treasury ladder of any size, and TIPS (`SCHP`) inherit it — Treasury obligations, same exemption.

## Not fixable as a value

[...]

Issue #210, opened 2026-08-17
EQUITY_DEFAULT_YIELD is 2-4x too high for US equity and 25-39% too low for international


MattOrefice:
Split out of #191 / PR (disclosure). #191 shipped the **disclosure** only and deliberately changed no yield value — correcting fifteen invented numbers to fifteen differently-invented ones is not progress. This is the numbers problem, and it deserves its own decision.

`EQUITY_DEFAULT_YIELD = 0.018` (`src/location_config.py:61`) is the silent fallback for any sleeve absent from `SLEEVE_ASSUMED_YIELD`. It is not an edge case: **7 of 26 register rows** resolve through it, and on cases A/B/D it carries **20.4% of the rendered drag KPI** — including the single largest row, GAOSX ($10,833 global allocation fund, $48.89 benefit), whose `multi_asset` sleeve has no table entry.

Measured against the committed dividends/prices tables as of 2026-08-11 (trailing-365-day distributions / last committed close, no network):

| sleeve on the default | measured | vs 1.80% |
|---|---|---|
| us_large_growth | 0.42% | **-77%** |
| us_sector_tech | 0.44% | **-76%** |
| us_large_quality | 0.85% | -53% |
| us_large_core | 0.92% | -49% |
| us_small_value | 1.23% | -32% |
| us_sector_healthcare | 1.58% | -12% |
| us_large_value | 1.81% | +1% |
| emerging_markets | 2.25% | +25% |
| intl_developed | 2.49% | +39% |

Systematically **too high for US equity** and **too low for international** — a single scalar cannot serve both, which is the real finding. Note `multi_asset` and `target_date` (RFUTX, $76K — the largest household holding) are blends and fit neither direction.

Also from #191, for whoever takes this: four of the fifteen **table** entries are checkable and two are wrong by ~2x — `tips` 2.50% vs measured 5.03%, `real_assets_commodities` 1.50% vs 2.84%. `core_fi_treasury` (-3%) and `real_assets_reit` (-10%) are roughly right. The other eleven are not derivable from committed data at all.

Decisions this needs, none of which #191 makes:
1. one default, several by asset class, or **no** default (raise on an unlisted sleeve — the RAISE option, which has five confirmations elsewhere in this codebase)
2. whether a yield basis is declared, and which
3. whether the fifteen table entries get revised at the same time or separately

[...]

PR #285, opened 2026-08-22
Declare what income IS, separately from how much of it there is

### LINKED ISSUE #278: The register's tax axis is the sleeve taxonomy: EQUITY_SLEEVES encodes exposure, not tax character, so a fix keyed on it is wrong by 4x

MattOrefice:
Closes #278. Rendered drag KPI **$276 → $268.63**.

## Why a vocabulary and not four exceptions

The register's rate was

```python
income_rate = state_only if _is_federally_exempt(sleeve) else ordinary
```

**an expression with exactly two outcomes.** US Treasury interest is federally taxed and PA-exempt, which is a **third**, and a boolean has nowhere to put one. So that line changed whichever way the fix was framed — and once it is not binary, a character selector exists whether or not anyone names it. The only remaining choice was a *named* vocabulary or an unnamed chain of `if`s. **Exceptions were not the weaker option for that case; they were unavailable.**

The rendered consequence is **#283**: page 8 says *"Critically, Treasury interest income is exempt from state and local taxes"* as the stated reason VGIT was chosen, while page 14 charged it 25.07%.

The exception-count argument this issue was filed on is demoted, not deleted — it forecasts that exceptions will grow, which is soft. The boolean has no third state today.

---

### 1. Three facts per member, with `treasury` as the proof

```python
#                     federal income   PA taxes it?   gain
"treasury":          ("ordinary",      False,         "ltcg")
```

Treasury **interest** is exempt from PA tax. A capital **gain** on a Treasury fund is an ordinary capital gain and PA taxes it. **Collapse the three into one value and either the interest is over-taxed or the gain is under-taxed** — the same failure the boolean had, one level down.

`test_treasury_needs_all_three_facts` pins both halves; `test_the_income_and_gain_facts_are_independent_across_the_table` is its non-vacuity check — if income and gain always agreed, the third slot would be ceremony, so it asserts two members disagree in each direction and that the two axes do not select the same set.

### 2. `FEDERALLY_EXEMPT_SLEEVES` deleted, not kept alongside

**Two places saying "muni" is the fourth-copy mechanism** — the one that produced four copies of a sentence in #228 and two contradicting statements in #284, on this exact subject. Keeping the old set as "the boolean case" would have been the cheap, wrong move.

Its two consumers went to different homes: the rate to `SLEEVE_TAX_CHARACTER`, the eligibility to its own set. `test_the_deleted_set_left_no_second_place_saying_muni` sweeps for it **through `ast`** rather than by grepping — the first version searched text and flagged `household.py`, because the comment *recording* the deletion quotes the retired expression. The checker found its own documentation, which is #282's shape exactly; a comment is not in the AST.

### 3. Eligibility separated, with its own reasons

A muni is wrong to shelter because of **what the destination wrapper does to it** — a pre-tax shelter converts exempt interest into ordinary income at withdrawal — which is not a property of the income's character. An **MLP** would be wrong for an unrelated reason (UBTI inside an IRA). Deriving eligibility from character forces the next entry to pretend to be a tax character.

That bundle is precisely why a state-exempt mirror was impossible: **a Treasury needs the rate half and must NOT be suppressed**, since there is still federal tax to save by relocating it. `test_a_state_exempt_sleeve_is_not_suppressed` pins the half the bundle could not express; `test_eligibility_is_not_derived_from_character` asserts the mechanism rather than the membership, because the two sets *coincide today* and membership could not tell them apart.

### 4. `section_1256` rejected, with the reason recorded

**It is the member someone will want to add.** 60/40 governs the **fund's internal gains**, not the character of what it distributes — those reach a holder as ordinary income (JHEQX is `ordinary`). Declaring it would encode a real tax rule at the wrong level. `test_section_1256_is_rejected_with_its_reason` asserts both that it is absent *and* that the reasoning is findable, because an unexplained absence gets filled in.

---

## The departure from #210's precedent, and why

`format_assumed_yield` leaves its `table` case **unmarked**, "because it is the only one a reader can verify unaided." `format_tax_character` marks **every** character.

**No character is verifiable unaided** — a reader cannot tell 25.07% from 20.67% by looking at a dollar figure. So there is no case that earns going unmarked, and an unmarked row would read as *"no character was declared here"* rather than *"the ordinary one"*, inverting the meaning. Following the precedent literally would have been the wrong call; the reason is in the docstring so nobody restores consistency later.

## `muni_in_state`: declared and unused

Nothing held is one. It stays because it is **the coordinate proving the federal and state axes are independent** rather than accidentally aligned — without it `treasury` reads as a special case with nowhere for the next state-exemption question to go. One line. `test_muni_in_state_is_declared_and_unused` asserts the deliberateness in both directions.

## The realization term

`cost_to_realize` was a flat `embedded_gain * ltcg` — right for almost everything, wrong for a physical-metal grantor trust at 28% federal. **One vocabulary, two rate functions**: declaring an asset twice would need a reconciliation rule between the declarations.

**RECORDED: this correction is entirely LATENT.** Gold cannot reach a taxable realization at all, and structurally rather than by today's positions — A/B require `annual_benefit > 0` and gold's authored yield is `0.0`; D requires a priority-map rank and `real_assets_gold` is in neither; C is in-shelter so the cost is 0 by construction. It moves no rendered figure. `test_gold_cannot_reach_a_taxable_realization_today` asserts the latency **so the day it ends is visible**.

## Disclosure

**Rendered, on the real book:**

> **Not every holding is taxed at the rates above.** What a holding pays depends on its income's **tax character**, declared per sleeve rather than inferred. Of 26 rows: **14** taxed as **ordinary income**; **6** taxed at the **qualified-dividend** rate; **3** taxed at the **REIT §199A** rate (ordinary less the 20% deduction); **2** **US Treasury obligations**, whose interest PA does not tax — the federal rate only; **1** a **collectible**, which changes the rate on a *sale* rather than on income. **4** of the 26 are not sized at all, so no income rate reaches them — their character still governs a sale. **Character is declared, not measured** — like the sleeve yields below, it is an authored fact about how income is taxed, and a sleeve whose character has not been declared **raises rather than defaulting to ordinary**.

**Counts reconcile to the register, not to a table in a design note:** 14 + 6 + 3 + 2 + 1 = 26 rows, 4 unsized. They are derived from the frame, never written down — `test_the_note_counts_are_derived_from_the_register_not_written_down` plus `test_the_note_moves_with_the_data` as its assert-it-mutated half.

**A correction to my own Part 2 table**, since this is exactly the reconciliation that was asked for: I tabulated ordinary as 12 and qualified as 4, treating the four `not_modelled` rows as their own bucket. The register says **14 and 6** — PDBC and GAOSX are `ordinary`, IDGT ×2 are `qualified`, and they are *also* unsized. Character and sized-ness are different questions, which is the design; my table conflated them. The implementation and the note were right; the design note's arithmetic was not.

**Per-row column**, rendered beside Assumed Yield:

```
Symbol             Sleeve Assumed Yield          Tax Character  Annual Benefit ($)
  JEPQ      Hedged Equity         6.00%               ordinary               23.26
  USRT Real Assets — REIT         4.00%             REIT §199A                5.21
   IAU Real Assets — Gold         0.00% collectibles (on sale)                0.00
  IDGT           Thematic  not modelled              qualified                 NaN
```

The yield note's *"the rate is the policy above"* now reads *"…**as it applies to that holding's declared tax character**"* — the old phrasing was true only while there were two rates.

## Mutation proof — 10 mutants, all killed by a named victim

| | mutant | |
|---|---|---|
| **M1** | default an undeclared sleeve to `ordinary` instead of raising | KILLED |
| **M2** | collapse the three facts — `treasury` loses its gain treatment | KILLED |
| **M3** | re-bundle eligibility onto the character table | KILLED |
| **M4** | apply `income_rate` where `realization_rate` belongs | KILLED |
| M5 | delete `muni_in_state` | KILLED |
| M6 | leave `ordinary` unmarked, per the yield precedent | KILLED |
| M7 | hardcode the note's counts instead of deriving them | KILLED |
| M8 | give REIT the qualified rate rather than the §199A deduction | KILLED |
| M9 | restore `FEDERALLY_EXEMPT_SLEEVES` beside the vocabulary | KILLED |
| M10 | declare `section_1256` | KILLED |

**M4 is killed on the register's own `cost_to_realize`, not on `realization_rate`'s return** — a function test passes on a register that never calls it, and the defect was a call site. Since gold cannot reach that path (above), the kill uses a *yield-bearing* sleeve given the collectibles character, so the assertion runs through the real call site. `test_the_income_rate_is_not_what_reaches_cost_to_realize` names M4 directly: `collectibles` income is 25.07% and its gain 31.07%, so the mutant produces a plausible wrong number rather than an error.

## The second commit, and what only the full suite could see

The first version resolved **character before yield**, which pre-empted #210's raise for an unlisted sleeve — that raise names the sleeve, the symbol, and all three config sets; mine is narrower. **Six yield-raise tests went red on the full suite.** No targeted run could have caught it: they exercise the two raises separately and both were individually correct. Only their *order* was wrong.

`test_an_unlisted_sleeve_still_gets_the_YIELD_raise_not_the_character_one` pins it now. Both raises are right for their own case — an unknown sleeve is a yield-table gap, an undeclared character is a sleeve the yield table already knows.

The same commit fixed `test_yield_proxies`, which derived its expected benefit from a hardcoded combined-ordinary rate. Right only while every sleeve got that rate — its fixture's sleeve is `us_large_core`, whose income is qualified — so left alone it would have asserted the defect this PR removes. It now derives the rate from the row's declared character.

## Verification

- **Collection:** 2047 selected (2087 − 40 deselected)
- **Guards UP**, personal mode, plain `pytest` at `a4b6c6b`: **13 failed / 2007 passed / 27 skipped** — 13 + 2007 + 27 = **2047**, reconciles
- The 13 are #187's twelve mode-sensitive render tests plus #177, both deliberate
- Rendered in local Streamlit before push; `demo.db` fingerprint identical across all 13 tables, `tracker.db` sha256 unchanged

---

# For the record — three things this PR established beyond its own change

## 1. Live templating caught arithmetic in an APPROVED design

The Part 2 design was reviewed and **approved with the wrong numbers in it**. It tabulated the character split as **ordinary 12 / qualified 4**, treating the four `not_modelled` rows as their own bucket. The register renders **14 / 6**: PDBC and GAOSX are `ordinary`, IDGT ×2 are `qualified`, and they are *also* unsized.

Character and sized-ness are **different questions** — which is the design, and the design note conflated them anyway. Nobody caught it at review, including the person who wrote it.

**That is the argument for templating over transcribing, stated as a concrete cost.** A hand-written count in the disclosure would have shipped the design note's error onto the page, where it would have contradicted the table directly beneath it, and no test would have failed — the number would simply have been wrong in a rendered sentence about a table a reader can see. The counts are derived from the frame precisely so the artifact cannot disagree with itself, and the check that confirmed them was asked for against the *register* rather than against the table, which is why it worked.

Pinned by `test_the_note_counts_are_derived_from_the_register_not_written_down` and `test_the_note_moves_with_the_data`.

## 2. A specified kill that was structurally impossible, reported rather than substituted

The mutation proof was specified to kill M4 on **IAU's rendered cost-to-realize**. That kill cannot exist: gold cannot reach a taxable realization at all — A/B require `annual_benefit > 0` and gold's authored yield is `0.0`, D requires a priority-map rank and `real_assets_gold` is in neither map, C is in-shelter so the cost is 0 by construction. A test pinning IAU's rendered cost would have been pinning a **zero**, and would have passed with the fix reverted.

Saying so beats finding a kill that *looks* like the specified one. The mutant is killed at the **call site** instead — a yield-bearing sleeve given the collectibles character — which is the property that mattered (`realization_rate` is reached from the register, not merely defined).

**The asserted latency is the part that matters more than the fix.** `test_gold_cannot_reach_a_taxable_realization_today` makes the day this state ends **visible rather than silent**: if gold ever gains a yield or enters a priority map, a collectibles row reaches a rendered `cost_to_realize` and the correction stops being latent. That is a change worth noticing, not a test worth deleting — the point of the assertion is the day it fails.

## 3. A defect with no single wrong line

The ordering bug had **no wrong statement anywhere**. `_tax_character`'s raise is correct. `_assumed_yield_with_source`'s raise is correct. Both were individually tested and individually green. Only their **order** was wrong, and the cost was a silent downgrade of the diagnostic a reader gets for an unlisted sleeve.

No targeted run could find it, and not by accident — a targeted run exercises the two raises separately, and separately they are both right. The fault lived in the relationship between them.

**Third instance this week of the full suite finding what a targeted run cannot**, and the common feature is the same each time: the fault is in a *relationship* rather than in a *statement*.

- #281 — a premise-check duplicated in a second file, structurally unreachable from the edit site
- #282 — a sweep blind to the wrapped form, invisible until a mutant supplied it
- #285 — two correct raises in the wrong order

A statement can be checked where it is written. A relationship can only be checked where both ends are in scope, which is the whole suite.


MattOrefice:
**Guards-DOWN convergence arm**, same tree `a4b6c6b`: **13 failed / 2007 passed / 27 skipped**, 2047 selected.

PR #287, opened 2026-08-22
VGIT's rationale: Pennsylvania, and stop calling it high-tax

### LINKED ISSUE #283: Page 8 renders a stale DC jurisdiction on VGIT's Treasury-exemption rationale — a DATA fix (seeder is insert-only; text lives in tracked demo.db and deploys publicly)

MattOrefice:
Closes #283. **A data PR with a prose payload** — the label matters, see below.

## The four-item check, re-run rather than trusted

The scoping was days old and this project's record on stale premises is poor, so all four were re-verified on this tree:

| | check | result |
|---|---|---|
| 1 | does page 8 read the text from a **file or a table**? | **table** — `pages/8_Research.py` selects `holding_rationale` from `securities` |
| 2 | is the seeder **insert-only**? | **yes** — two `if existing is None:` guards, two `INSERT INTO securities`, **zero** `UPDATE`/`ON CONFLICT` |
| 3 | is the DB **tracked**? | **yes** — `data/demo.db` is in `git ls-files data` |
| 4 | is page 8 **mode-gated**? | **no** — its only `if IS_DEMO:` is `st.info(get_demo_banner_text())`, a banner, and `app.py` routes it unconditionally |

**A correction on item 4.** The earlier report printed *"[no output above = renders publicly]"* when the grep **did** have output (`159:if IS_DEMO:`). The conclusion was right and the evidence line was a canned echo that did not read its own result — exactly the shape this check exists to catch. Line 159 is now read, not assumed.

## The approved wording was wrong, and checking it is what found that

The wording approved was a straight **DC → PA** swap. That trades a wrong jurisdiction for a wrong **characterisation**: PA's flat **3.07%** is one of the lowest state income taxes in the country, so *"a high-income-tax jurisdiction"* is true of DC and **false of PA**.

Three claims have to hold at once, and the swap fixes one:

| claim | before | after |
|---|---|---|
| the exemption | **correct** (and contradicted by the register until #278) | kept |
| the jurisdiction | wrong — DC | **Pennsylvania** |
| the magnitude | *"high-income-tax jurisdiction"* — false for PA | *"a real, if modest… at Pennsylvania's flat rate"* |

**No rate is stated.** `TAX_PROFILE` holds 3.07%; putting it in seed prose mints a second copy of a config value, which is the mechanism behind four copies of one sentence (#228) and two contradicting statements (#284). *"Real, if modest"* carries the size without the number — at VGIT's 4.00% assumed yield the advantage is ~12bp/yr.

## Rendered, before and after

From an actual page-8 render, personal mode:

**Before**
> Critically, Treasury interest income is exempt from state and local taxes — **a real after-tax advantage in a high-income-tax jurisdiction like DC** over investment-grade corporate bond funds with similar yield.

**After**
> Critically, Treasury interest income is exempt from state and local taxes — **a real, if modest, after-tax advantage at Pennsylvania's flat rate** over investment-grade corporate bond funds with similar yield.

## The migrations

Following the `migrate_demo_vea_rationale.py` / `migrate_personal_vea_rationale.py` **pair** — the exact precedent: same change, both books, two scripts, each targeting its DB explicitly and never `get_db_path`. Both expose `main()` only, so `bootstrap.run_pending_migrations` does not auto-run them.

**Two deliberate differences from the template:**

1. **A targeted substring swap**, not a whole-string overwrite — so prose the script did not author is preserved byte-for-byte. (The demo template copies the whole seed string in; the personal one already does the swap. Taking the swap for both keeps the pair identical in mechanism.)
2. **It asserts `cursor.rowcount == 1` and rolls back otherwise.** **No existing migration in `tools/` checks rowcount** — and a rationale `UPDATE` whose `WHERE` matches too widely **succeeds silently**. It would surface only in a rendered diff, on a tracked binary, after the commit. The ticker predicate makes a wide match unlikely; the assertion is what makes it impossible to ship unnoticed.

Registered as **step 5** in `tools/sync_personal_prose.py`. Four sibling docstrings said *"sequences all four"* — adding a fifth made four stale copies at once, so they now state **no count**. A count in prose is a second copy of a fact the list already holds.

## demo.db: asserted, not fingerprinted

The fingerprint diagnostic necessarily changes here, so it was replaced with assertions against a pre-migration snapshot:

```
table LISTS identical: True (13 tables)
tables whose CONTENT changed: 1
  securities: rows 27 -> 27   hash 5227bfea8c4a -> 6d3d5a58ac1d

ASSERTION 1: exactly ONE table changed (securities), row count unchanged — PASS
ASSERTION 2: securities rows differing: 1 -> ['VGIT']                      — PASS
ASSERTION 3: the delta is EXACTLY the intended substring swap              — PASS
  chars 1085 -> 1083   (prose outside the swap is byte-identical)
ASSERTION 4: columns changed on VGIT: ['holding_rationale']                — PASS
```

Post-commit, `fingerprint_db` reads IDENTICAL against HEAD again, which is the check that the committed binary matches the working tree.

## tracker.db: what was done, not a pass/fail

Gitignored, so it does not travel with the repo and must be run per machine via the documented entry point. Migrated through `tools/sync_personal_prose.py`, which reported `1 applied, 4 already-applied, 0 aborted`.

```
before : b5e9298a589e6daddf6dab3ab2bf1654fb8891609854822ab5629eb5e6249502
after  : 87286a072300e598e09a286406d631e7e9a350435c186de9b1d4918ff09f51fa
CHANGED: True
```

Changed **for the intended reason**: `Pennsylvania` + `flat rate` present, `like DC` gone, **0** rows still mentioning DC, securities row count unchanged at 74. Reporting UNCHANGED here would have meant the migration did not run.

## Mutation proof — 11 mutants, all killed by a named victim

| | mutant | |
|---|---|---|
| **M1** | revert the source string to the DC wording | KILLED |
| **M2** | **widen the migration's `WHERE`** to `LIKE '%like DC%'` | KILLED |
| M3 | drop the rowcount assertion (the silent-success guard) | KILLED |
| M4 | whole-string overwrite instead of a substring swap | KILLED |
| M5 | drop the abort on unexpected text | KILLED |
| M6 | break idempotence | KILLED |
| M7 | state the tax rate in the prose | KILLED |
| M8 | name PA but keep "high-income-tax" | KILLED |
| M9 | unregister the personal migration | KILLED *(survived first)* |
| M9b | run a migration the docstring does not document | KILLED |
| M10 | restore a hardcoded migration count in a sibling | KILLED |

**M9 survived the first pass**, and the instrument was at fault. The registration test searched the runner's **source text** for the module name — and the module **docstring** names it too, as step 5. So deleting it from `_MIGRATIONS` left the string present and the test green: **a positive source assertion satisfied by prose promising the thing.** The runner would have documented a migration it no longer runs.

It now asserts `runner._MIGRATIONS` directly, with a second test (M9b's victim) pinning that the docstring documents everything the list runs — because those are two copies of one fact and either can drift.

**M8 is the one that pins the review finding**: naming PA while keeping "high-income-tax" is the change that would have shipped under the originally-approved wording.

## Verification

- **Collection:** 2067 selected (2107 − 40 deselected)
- **Guards UP**, personal mode, plain `pytest` at `776e1e6`: **13 failed / 2027 passed / 27 skipped** — 13 + 2027 + 27 = **2067**, reconciles
- The 13 are #187's twelve plus #177, both deliberate
- `demo.db` fingerprint IDENTICAL against HEAD post-commit


MattOrefice:
**Guards-DOWN convergence arm**, same tree `776e1e6`: **13 failed / 2027 passed / 27 skipped**, 2067 selected.

Books after both arms: `demo.db` fingerprint IDENTICAL against HEAD across all 13 tables — i.e. the committed binary matches the working tree, which is the post-migration form of that check. `tracker.db` sha256 unchanged from its post-migration value, so the suite did not move it. Tree clean, guards restored to UP.

Issue #291, opened 2026-08-22
Nothing checks that a comment pointer still resolves — 278s ast sweep is blind to comments by construction, which hid a stale FEDERALLY_EXEMPT_SLEEVES reference for three PRs


MattOrefice:
Named while fixing #286. It is a gap **created by a fix**, not one that was always there.

## What happened

#278 deleted `FEDERALLY_EXEMPT_SLEEVES`. Its sweep — `test_the_deleted_set_left_no_second_place_saying_muni` — was written to catch the symbol coming back. The first version searched **file text** and flagged `household.py`, because the comment *recording* the deletion quotes the retired expression. The checker found its own documentation.

The fix was to read through **`ast`**, so only real code is examined. That worked, and it is the right fix for that problem.

**It also made the sweep structurally blind to comments** — and two comments in `location_config.py` were still pointing at the deleted symbol, unnoticed for three PRs:

```python
# ... its exemption lives in
# FEDERALLY_EXEMPT_SLEEVES below, not as a fake zero here.     <- the set is not below; it is gone

"high_yield_muni": 0.040,  # ... see FEDERALLY_EXEMPT_SLEEVES  <- points at nothing
```

Both are corrected in #286's PR. The gap is that **nothing would have caught them, and nothing would catch the next one.**

## The shape

**A stale comment pointer is a claim that a symbol resolves.** It is not decoration: `see FEDERALLY_EXEMPT_SLEEVES` sends a reader to look for something that does not exist, and their most likely conclusion is that they are looking in the wrong place rather than that the comment is wrong.

The AST fix and this gap are two sides of one trade:

| | sees code | sees comments |
|---|---|---|
| text search | yes | yes — including its own documentation |
| AST | yes | **no** |

Neither is right alone. The AST sweep is correct about *code* and silent about *prose*; a text sweep is correct about *prose* and cannot distinguish a live reference from a historical one.

## The check — #288's shape applies

> **assert that no comment names a symbol that no longer resolves**

Concretely: extract identifiers from comments and docstrings under `src/`, filter to names that look like module-level constants (`ALL_CAPS_WITH_UNDERSCORES`), and assert each either resolves somewhere or sits in a context marked as historical.

[...]

Issue #290, opened 2026-08-22
BLOCKED on #289: is a covered-call payout modellable as a yield at all? 40 percent of the drag KPI


MattOrefice:
Split out of #289 deliberately. That issue is about **coverage** — how much of the KPI is undeclared. This is a **modelling decision**, and it should not be buried inside a coverage sweep because it is the decision actually worth making.

## The question

`hedged_equity` carries an authored **6.00%** yield and produces **$107.62** of the $268.62 drag KPI — **40% of the total**, and 89× the entry #286 just fixed.

The holdings are JEPI, JEPQ, JHEQX (and HELO in case C). Their payout is **largely option premium and return of capital**, not income in the sense the yield table models. A covered-call or ELN-based fund distributes what it harvested from selling volatility; the character and the amount both move with the option market rather than with an underlying coupon or dividend stream.

**So the question is not "what number should this be". It is whether the sleeve is modellable at all** — a proxy candidate or a `NOT_MODELLED_SLEEVES` candidate.

## Why this is #210's argument arriving on a much larger sleeve

#210 moved `real_assets_commodities` to `NOT_MODELLED_SLEEVES` on exactly this reasoning: PDBC distributes **realized gains** rather than income, so a trailing-twelve-month yield measures what it happened to sell that year — measured at 0.97 / 1.70 / 0.00 / 40.38 / 13.15 / 4.18 / 4.41 / 2.84% across successive year-ends. The shipped 1.50% and the proposed 2.84% were both single-year snapshots of that range.

`hedged_equity` has the same structural property — a payout driven by harvested premium rather than by an income stream — on a sleeve carrying **$107.62 of drag against `real_assets_commodities`'s $0**. The precedent was established on the smallest possible stake, and the argument, if it holds, applies to the largest.

**That asymmetry is the reason to decide it deliberately rather than let it fall out of a coverage sweep.** Refusing to size 40% of the KPI is a much larger statement than refusing to size 0% of it, and it should be made on its merits.

## What would have to be true either way

**To proxy it** — the payout has to be stable enough that one year's figure is informative about the next. That is precisely the claim #210 rejected for commodities, and it needs testing here rather than assuming. JEPI, JEPQ and JHEQX are **not in the price cache at all** (0 rows each), so measuring even one year requires the cache-scope decision in #289.

[...]

Issue #289, opened 2026-08-22
96.1% of the drag KPI rests on undeclared yield bases; hedged_equity alone is 89x the entry #286 fixed


MattOrefice:
Split out of #286, and **it is the primary issue** — #286 and #231 are both rounding errors inside it.

## 96.1% of the rendered drag KPI rests on an undeclared basis

Measured on the 2026-08-22 book, A/B/D rows only:

| sleeve | declaration | yield | drag | held |
|---|---|---|---|---|
| `hedged_equity` | **AUTHORED — none** | 6.00% | **$107.62** | JEPI, JEPQ, JHEQX |
| `multi_sector_fi` | **AUTHORED — none** | 4.00% | **$65.85** | FIWDX, GBOSX |
| `floating_rate` | **AUTHORED — none** | 5.50% | **$43.08** | BFRIX |
| `core_fi_credit` | **AUTHORED — none** | 3.50% | $21.16 | JCPB |
| `real_assets_reit` | **AUTHORED — none** | 4.00% | $15.53 | USRT, VNQ |
| `emerging_markets` | proxy | 1.70% | $9.28 | |
| `liquid_alt` | **AUTHORED — none** | 2.00% | $3.61 | BILPX |
| `core_fi_treasury` | proxy *(since #286)* | 3.97% | $1.20 | VGIT |
| `tips` | construction | 4.61% | $0.92 | SCHP |

**$258.06 of $268.62.** `hedged_equity` alone is **89×** the entry #286 just fixed. Three sleeves — `hedged_equity`, `multi_sector_fi`, `floating_rate` — total **$216.55, 81% of the KPI**, and state nothing at all.

Every one of them traces to `6cd6f99` (2026-06-01), where **fifteen of fifteen entries were exact multiples of 0.5%**. Authored in one sitting. Of that block, three have since been examined and **three moved**: `tips` +84%, `real_assets_commodities` to unmodellable (measured range 0.00%–40.38%), `high_yield_muni` −11%. Twelve remain.

## The chain-selection finding

> **#231 named `tips` because it was the best-documented entry ($0.03). #286 named `core_fi_treasury` because its rationale happened to state a duration ($0.01). The money is in `hedged_equity` + `multi_sector_fi` + `floating_rate` — $216.55, 81% of the KPI — which state nothing and so offered nothing to notice.**

[...]

Issue #231, opened 2026-08-17
TIPS yield uses the 10-year point; SCHP holds broad-maturity TIPS ($0.03/yr post-#278 — see #286 for the larger, undeclared instance)


MattOrefice:
The `tips` entry in `SLEEVE_ASSUMED_YIELD` is constructed as `DFII10 + T10YIE` (#213) — the **10-year** real yield plus the **10-year** breakeven.

SCHP tracks a broad-maturity TIPS index, so a duration-matched real yield and breakeven would be more exact than the 10-year point. The approximation is declared in `SLEEVE_YIELD_CONSTRUCTION` rather than left implicit.

**Why it was not fixed in #213:** finding the right tenor is a measurement question, not a wording one — it needs SCHP effective duration, then whichever FRED series bracket it, then a check that both legs are available at the same tenor (breakevens are published at 5y/10y/30y; TIPS real yields at 5y/7y/10y/20y/30y, so a duration-matched *pair* may not exist and the construction may have to interpolate or stay at 10y by necessity).

**Why it is low priority:** measured at #213, the entire 2.50%-9.88% range moves the Asset Location drag KPI by \$1.68 on a \$90.68 holding — 0.17% of the total. A tenor refinement is a fraction of that. Latent, not wrong: the entry declares the approximation.

MattOrefice:
## The failure mode is staleness nobody notices, not the tenor

Worth separating from the tenor question, because it is the likelier way this entry goes wrong.

The construction is **frozen at 2026-07-20** and the entry says so — a dated snapshot that announces its date is honest, and that was a deliberate choice over a live fetch (the yield table has no network dependency, and `test_the_construction_is_frozen_not_fetched` pins that it keeps none). Four weeks of vintage is a fair price for that property.

**But nothing currently prompts a refresh.** There is no reminder, no staleness assertion, no test that reddens as the as-of date recedes, and the rendered note discloses the *nature* of the figure ("a dated snapshot, not a live feed") without disclosing that this particular snapshot is now old. So the realistic failure is not that the 10-year point approximates broad-maturity badly — it is the entry sitting at a year stale while every surface still reads as correct, because it *is* correctly labelled. It would take a reader opening `SLEEVE_YIELD_CONSTRUCTION` and doing the subtraction.

[...]

PR #292, opened 2026-08-22
core_fi_treasury declares a basis: IEF TTM, 3.97%

### LINKED ISSUE #286: core_fi_treasury's yield is authored 4.00% with no declared basis, against a duration the repo states on page 8 — 3x #231 and undocumented
Found while scoping #231, which turns out to be the smaller and better-documented instance of this.

## The defect

`SLEEVE_ASSUMED_YIELD["core_fi_treasury"] = 0.040` — **authored, with no declared basis at all.** It is not in `SLEEVE_YIELD_PROXY` and not in `SLEEVE_YIELD_CONSTRUCTION`. Nothing records where 4.00% came from, what instrument it describes, or as of when.

**The repo states the duration that should have set it, and renders it to a reader.** VGIT's holding rationale, on page 8:

> Duration is modestly shorter (**~5.5 years** vs IEF's ~7.5 years), appropriate for a 6% sleeve inside a 78% growth portfolio.

The nominal Treasury curve at the TIPS construction date (2026-07-20, from `macro_cache`):

```
DGS5   4.33%
DGS7   4.45%      -> 5.5y interpolates to 4.36%
DGS10  4.60%
DGS20  5.12%
DGS30  5.11%
```

**A duration-matched figure is ~4.36%. The entry says 4.00% — a 36bp gap, and the authored value UNDERSTATES.**

## Sized

| | |
|---|---|
| VGIT today | $137.02 × 4.00% × 22.00% = **$1.21** |
| At the 5.5y nominal (4.36%) | $1.31 |
| **Gap** | **$0.11/yr** |

Small in dollars because the position is $137. The **rule** misprices a Treasury ladder of any size, and the rate is now correct (`treasury` character, #278) so the yield is the only remaining wrong multiplicand.

Compare #231, which raises the same defect for `tips`: **15bp, worth $0.03/yr.** This one is **three times larger and has no declared basis**, where the TIPS entry declares its approximation in `SLEEVE_YIELD_CONSTRUCTION` and names #231 in its own note.

## The general finding, which is the reason this is filed separately

> **TIPS was noticed because it is the only entry documented well enough to notice.**

`tips` records its formula, its two component series, an as-of date, a Fisher-identity cross-check against `DGS10`, and an explicit paragraph headed *"APPROXIMATION, DECLARED"* naming the tenor problem and sizing it. All of that is why #231 exists.

`core_fi_treasury` records **nothing** — so there was nothing to read, nothing to check, and no issue got filed against it for months.

**Documentation makes defects findable, so the best-declared entry attracts the issue while the undeclared one sits unexamined.** That is an argument for suspecting the entries that declare nothing, and it inverts the intuition that a documented approximation is the riskier one. A declared approximation is bounded and someone is watching it; an undeclared value is unbounded and nobody is.

The practical form: **when an issue names one entry in a table, check the entries that declare least, not the ones that declare most.**

## Interaction with #278

This is the same holding as #283, one axis over. #278 fixed VGIT's **rate** (`treasury` character — federal only, PA does not tax Treasury interest). #283 is about the **prose** describing that exemption. This is about the **yield** the rate multiplies.

Three defects on one $137 position, on three different axes, each found by a different route.

## What it needs

1. **Decide whether the entry gets a declared basis at all**, and of what kind. `SLEEVE_YIELD_PROXY` (a benchmark's TTM distribution yield) is the wrong shape here — a Treasury fund's distribution yield is not its yield-to-maturity. `SLEEVE_YIELD_CONSTRUCTION` is the closer precedent: a duration point on a published curve, recorded with its series and as-of.
2. **Whether 4.00% was a deliberate haircut for something** — a distribution-yield-versus-YTM distinction, a rounding, a conservative choice — or simply never revisited. The entry declares no basis, so there is nothing to read and this is unanswerable from the repo.
3. Note the constraint #231 hit: the yield table deliberately has **no network dependency** (*"FROZEN AND DATED, NOT LIVE … That property is worth more than the vintage"*). Unlike #231's case, the series this one needs (`DGS5`/`DGS7`) **are already cached**, so a construction here is available where a duration-matched TIPS pair is not.

## Related

- #231 — the same defect on `tips`: smaller, declared, and the reason this was found
- #278 — made the rate correct, leaving the yield as the only wrong multiplicand
- #283 — the prose describing the same holding's exemption

MattOrefice:
Closes #286. Drag KPI **$268.63 → $268.62**.

## Read this first: it is the second-smallest measurable entry, and it is not progress on the problem

**#289 is the issue that matters.** 96.1% of the rendered drag KPI still rests on undeclared bases, and `hedged_equity` alone is **89×** the entry this PR fixes. Three sleeves — `hedged_equity`, `multi_sector_fi`, `floating_rate` — total **$216.55, 81% of the KPI**, and declare nothing.

This PR moves **one cent**. It is worth doing for the reason below, not for the money, and it should not be read as the undeclared-basis problem being worked.

## #286's own sizing was wrong, and the sign flips

The issue set the authored 4.00% against the **nominal yield curve** at VGIT's ~5.5y duration (4.36%) and called it a 36bp understatement worth **+$0.11/yr**.

But this table's declared basis — for every proxied entry — is a **trailing-twelve-month distribution yield**: coupon actually paid out, not a point on the curve. A fund holding older, lower-coupon bonds distributes *below* current yield-to-maturity, so the two are different quantities.

| | |
|---|---|
| IEF TTM distribution yield @ 2026-08-11 | **3.97%** ($3.6830 over 12 payments, price 92.87) |
| VGIT, the held ticker | 3.90% |
| authored | 4.00% |
| **error** | **3bp, OVERSTATED** |
| impact | **−$0.01/yr** |

39bp from what the issue claimed, and on the other side of the authored value. `test_a_distribution_yield_is_not_a_curve_point` pins this so the wrong comparison cannot be re-derived.

**Same correction #286 applied to #231, now applied to #286.**

## Why do it anyway

The entry arrived at `6cd6f99` (2026-06-01) inside a block whose **fifteen entries are all exact multiples of 0.5%**, with no comment then or since. Authored in one sitting; never measured. That is a different fix from measured-and-gone-stale — a first measurement, not a refresh.

The base rate for that block: three entries have since been examined and **three moved**.

| | 2026-06-01 | now | |
|---|---|---|---|
| `tips` | 0.0250 | 0.0461 | **+84%** |
| `real_assets_commodities` | 0.0150 | — | **unmodellable**, measured 0.00%–40.38% |
| `high_yield_muni` | 0.0450 | 0.0400 | −11% |

Twelve remain authored. **Being close by luck and being measured are indistinguishable from the value alone** — only the measurement separates them, and this entry turned out to be one of the lucky ones.

**IEF and not VGIT**, per the map's `BENCHMARK THROUGHOUT` convention. Taking the held ticker on one row is held-weighted reasoning applied selectively — the thing that made `intl_large_value` take EFV over AVIV at an +85% spread.

## Two things the existing harness taught mid-change

**1. A per-entry as-of silently fails the recomputation test.** The first draft gave this entry its own as-of (2026-08-19). `test_entry_values_match_their_proxies_recomputed_from_the_cache` resolves **every** proxy at one pinned date (2026-08-11) — so the entry recomputed to 0.0397 while I had written 0.0394, and it went red. The map's as-of is shared, and now says so.

**2. `EXPECTED_PROXIES` could not see a new entry.** It is a test-side duplicate of the config map, and every assertion iterated the **duplicate** — so a sleeve added to the real map was invisible: 13 pinned, a 14th unchecked, nothing red. Now asserted **both ways**.

> One-directional coverage of a mirrored enumeration is `assert-collections-are-complete` with the collection split across two files.

## Also: two dangling pointers from #278

`FEDERALLY_EXEMPT_SLEEVES` was deleted, and two comments still pointed at it — *"its exemption lives in `FEDERALLY_EXEMPT_SLEEVES` **below**"* and *"see `FEDERALLY_EXEMPT_SLEEVES`"*. Both now name `SLEEVE_TAX_CHARACTER`.

**That gap was created by a fix.** #278's sweep reads through `ast` specifically so it could not flag its own documentation — and the same choice made it blind to genuine stale *comment* pointers. Filed as **#291**; the check has to separate a **pointer** (`see X`, `lives in X`) from a **record** (`REPLACES X`), because this repo deliberately keeps corrections that name deleted things.

## Mutation proof — 9 mutants, all killed by a named victim

| | mutant | |
|---|---|---|
| **M1** | revert to the authored round 0.040 | KILLED |
| **M2** | use the nominal **curve point** (0.0436) instead of the distribution yield | KILLED |
| M3 | take the held ticker (VGIT) instead of the benchmark | KILLED |
| M4 | drop the entry from the proxy map — undeclared again | KILLED |
| M5 | drop the spread from the entry comment | KILLED |
| M6 | drop the held-coverage from the entry comment | KILLED |
| M7 | erase the never-measured provenance | KILLED |
| M8 | restore a dangling pointer to the deleted set | KILLED |
| M9 | restore the one-directional test enumeration | KILLED |

M1 and M2 are the pair that matter: one reverts to the assumption, the other substitutes the wrong *quantity* — and the second is the mistake the issue itself made.

## Verification

- **Collection:** 2077 selected (2117 − 40 deselected)
- **Guards UP**, personal mode, plain `pytest` at `8466fc3`: **13 failed / 2037 passed / 27 skipped** — 13 + 2037 + 27 = **2077**, reconciles
- The 13 are #187's twelve plus #177, both deliberate

## Filed alongside

- **#289** — 96.1% of the KPI on undeclared bases. **The primary issue.** Carries the chain-selection finding: #231, #286 and #289 were each chosen by an incidental hook rather than by size, which is my own recorded rule failing on the session that recorded it.
- **#290** — whether a covered-call payout is modellable as a yield at all. #210's argument arriving on a sleeve 89× larger than the one that established it. Filed separately because it is the decision worth making, not a coverage item.
- **#291** — nothing checks that a comment pointer still resolves.


MattOrefice:
**Guards-DOWN convergence arm**, same tree `8466fc3`: **13 failed / 2037 passed / 27 skipped**, 2077 selected.

Books after both arms: `demo.db` identical across all 13 tables; `tracker.db` sha256 unchanged. Tree clean, guards restored to UP.

Issue #213, opened 2026-08-17
tips at 2.50% is wrong in kind, not stale — and no single measured number is defensible


MattOrefice:
Split out of #210, which deliberately changes no yield value. This one needs a decision of a different kind from every other entry in the table.

## It is not 2x stale, it is measuring the wrong thing

`SLEEVE_ASSUMED_YIELD["tips"] = 0.025` reads as a **real yield**. The location model needs **taxable income thrown off**, and for TIPS that includes the **inflation accrual** — principal adjustments that are taxable as they accrue and pay no cash until maturity. That accrual is the canonical reason TIPS belong in a shelter at all, and the phantom-income argument already appears in `saa_sleeves_taxable`'s authored prose. So 2.50% understates precisely the quantity this model exists to size.

## No single measured number is defensible

SCHP, five bases, one day (2026-08-11, price $25.90, committed data only):

| basis | figure |
|---|---|
| table | 2.50% |
| TTM as cached (10 of ~12 months) | 5.03% |
| TTM scaled for the two missing months | 6.04% |
| last 6 months annualised | 7.26% |
| last 3 months annualised | **9.88%** |
| calendar 2025 (8 months cached) | 3.71% |

A ~4x spread, and it is not noise — the monthly distribution roughly doubled from ~$0.09 in late 2025 to ~$0.21 by mid-2026 as the accrual rose. **No trailing window measures a forward-looking component**, so no TTM figure is the right answer either.

## What a defensible value would look like

A **constructed** figure with its construction declared — real yield + an assumed inflation rate — rather than a measured one. That is a *formula*, not a proxy, which makes it a different kind of assumption from every other #210 entry (those get a declared benchmark proxy). It is why it does not belong in #210's PR 2: mixing a formula-based entry into a proxy-based table without saying so would recreate the undeclared-basis problem #191 closed.

Decisions needed:
1. real yield from where — a declared source, or authored?
2. which inflation assumption — and is it the same one used anywhere else in the repo?
3. does the disclosure name the construction, given #191's rule that a declared basis must not hide its own sensitivity?

## Open, and not resolvable without a fetch

[...]

PR #296, opened 2026-08-22
A yield declares HOW it was arrived at, not where it is stored

MattOrefice:
Part of #289 — the Q5 half, which is **independent of the cache decision**. No sleeve is measured here; the table stops pretending some of its values were.

## The finding: a measured date is a fact about the world, an authored date is a fact about the author

**Only one of them decays.** A constructed entry drifts because the world moves — #231 measured a year of TIPS drift at a median 31bp, p90 76bp, 272bp in the worst 365-day window. **An authored value cannot drift, because there is nothing it could drift from.** A 6.00% judgement is exactly as justified in 2027 as on the day it was written.

Every other decision in this PR follows from that one sentence, which is why it is written into the config rather than left to be re-derived:

| | |
|---|---|
| **the key is `authored`, not `as_of`** | those two dates answer different questions, and a shared name invites comparing them |
| **the date is 2026-06-01, not today** | see below |
| **it renders in the Assumptions note, never beside the figure** | *"authored 82 days ago"* next to a number invites the inference that the value **drifted out of date** — wrong in a harmful direction, because the failure mode is that it may never have been right, and elapsed time neither creates nor cures that |
| **no review interval** | #231's cadence exists because the drift it guards was **measured**. Inventing one here would put a number with no basis inside the shape built to end numbers with no basis |

**Why 2026-06-01.** All twelve entries trace to **one commit — `6cd6f99`** — where fifteen yields arrived together and **every one was an exact multiple of 0.5%**. They were authored *wholesale*, in one sitting, and the shared date is evidence rather than bookkeeping. Stamping today's date would claim twelve reviews that did not happen: **declaring something authored is not reviewing it.**

The note says so, and derives it — the "all on the same day" clause fires only because `stamps[0] == stamps[-1]`, so a map with mixed dates renders a range instead and the evidence cannot drift from the sentence.

## The defect

`YIELD_BASES` held `table`, documented as *"the sleeve has an entry in `SLEEVE_ASSUMED_YIELD`"* — **a fact about storage.** It spanned three provenances that rendered identically:

```
BFRIX  floating_rate     AUTHORED       basis=table   renders "5.50%"
VGIT   core_fi_treasury  proxy          basis=table   renders "3.97%"
SCHP   tips              construction   basis=table   renders "4.61%"
```

**12 of 27 entries were authored, and all twelve are exact multiples of 0.5%.** The disclosure paragraph had counted proxied/constructed/authored since #213 — so the note and the per-row column disagreed about how many kinds existed.

## Rendered

```
Symbol             Sleeve                 Assumed Yield          Tax Character
  JEPQ      Hedged Equity              6.00% (authored)               ordinary
  USRT Real Assets — REIT              4.00% (authored)             REIT §199A
   IAU Real Assets — Gold 0.00% (no cash flow to yield) collectibles (on sale)
  IDGT           Thematic                  not modelled              qualified
```

> **15 of 26 rows** use an **authored** yield — a **judgement, not a measurement**. […] They were authored **2026-06-01** — all on the same day, wholesale rather than judged entry by entry. **That date is not a vintage.** An authored value does not go stale, because there is nothing it could drift from — the date says how long it has gone **unexamined**, not how far it has moved.
>
> **1 of 26 rows** has **no cash flow to yield** — bullion and crypto distribute nothing, so its zero is a fact about the asset rather than an estimate of its size. Marked separately from the authored rows because a zero labelled *authored* invites the next review to reconsider whether gold pays a dividend, and that question has a fixed answer.

## `structural` is a separate kind, and the render decided it against my own first draft

The draft folded structural into authored, arguing a kind must earn its name by changing behaviour and that this one changed none. **Rendering showed otherwise.** `IAU` is a **live register row** and came out as:

```
IAU   Real Assets — Gold   0.00% (authored)
```

which reads as *"somebody estimated zero"* where bullion has **no cash flow to yield**. The suffix says that rather than `(structural)`, because a zero labelled with a category name still reads as an estimate of zero.

**Two premises offered for this decision did not survive checking, and the conclusion did — on different evidence.** Recorded because I was asked to build on them:

- *"the resolver already distinguishes structural from judged by never consulting the table"* — **it does not.** All kinds take the same three-way branch and read the same table; gold's `0.000` comes *from* it.
- *"crypto and real_assets_gold never reach the register, so authored would have applied to three sleeves contributing \$0.00"* — **gold does reach it**, one case-C row, and it is the only row the marker is visible on. Crypto does not on this book; cash is excluded by name.

So the render decided it, but in the opposite direction from the one stated: **because gold renders, not because it does not.** The surviving argument is the third — the harm is to the **reviewer**, and it holds at any row count.

## The verifiability justification was wrong, not incomplete

`format_assumed_yield` said the `table` case is unmarked *"because it is the only one a reader can verify unaided."* Tested against the kinds it spanned, that does not track verifiability at all:

| kind | checkable unaided? |
|---|---|
| `constructed` | **yes** — formula and components are in the config; add 2.35 + 2.26 |
| `proxy` | no — needs the price cache and a TTM computation |
| `authored` | nothing to check |

`constructed` is the **most** checkable and would be the one left unmarked under the stated rule. The surviving distinction is **declared or not**.

**Not marking all four**, and this does not transfer from `format_tax_character`: there unmarked was genuinely ambiguous because *no* character is verifiable from a dollar figure. Here unmarked has one meaning once the undeclared kinds are marked, and suffixing 15 of 27 rows costs legibility for nothing. Recorded so a later consistency pass does not "fix" it.

## The reconciliation caught a dropped row

The note's counts template from the frame **and must sum to the row count** — the shape that shipped 12/4 against a real 14/6 in #285's design note. Adding `structural` **silently dropped a row from the prose** (16 authored → 15, with no structural clause) until `test_the_note_counts_reconcile_to_the_row_count` caught it.

One sentence had also gone false and is fixed: the note said the authored rows *"carry **no date at all**"* — true until this PR.

## Seventeen tests migrated, in a second commit, and only the full suite found them

Targeted runs were green at every step. **The full suite came back 30F against an expected 13F** — seventeen tests across five files pinning the old `table` literal or the `"no declared basis"` phrasing, none reachable from the files this change touched. **Third instance this week of a fault living in a relationship rather than a statement.**

Migrated, not edited to match:

- **fixtures** hardcoding `"table"` now **derive** the basis from the config maps, so they cannot drift the next time a sleeve changes kind
- the two `YIELD_BASES` pins (three → six) each say *what split and why*
- assertions requiring `"no declared basis"` are re-pointed at the property they protected — *the reader must be told the number was not measured* — because that phrase became **false in the good direction**

## Mutation proof — 12 mutants, all killed by a named victim

| | mutant | |
|---|---|---|
| **M1 / M1b** | fold structural into authored | KILLED **on the rendered column**, not the constant |
| **M2 / M2b** | stamp today's date; rename the key to `as_of` | KILLED |
| **M3 / M3b** | mark the proxied / constructed kinds too | KILLED **on the disclosure's own claim** |
| M4 | render the date beside the figure | KILLED |
| M5 | drop the structural clause from the note | KILLED (on the reconciliation) |
| M6 | hardcode the note counts | KILLED |
| M7 | restore "the authored rows carry no date" | KILLED |
| M8 | leave a sleeve undeclared | KILLED |
| M9 | restore the wrong verifiability justification | KILLED |

## Verification

- **Collection:** 2095 selected (2135 − 40 deselected)
- **Guards UP** at `d1daf1a`: **13 failed / 2055 passed / 27 skipped** — 13 + 2055 + 27 = **2095**, reconciles
- Failure set **byte-identical** to the expected 13 (#187's twelve + #177), `diff` empty


MattOrefice:
**Guards-DOWN convergence arm**, same tree `d1daf1a`: **13 failed / 2055 passed / 27 skipped**, 2095 selected.

Books after both arms: `demo.db` identical across all 13 tables; `tracker.db` sha256 unchanged. Tree clean, guards restored to UP.

MattOrefice:
## Zero-suppression pinned both ways (third commit, `a37263f`)

**No suppression needed shipping — it already worked.** All six clauses have been guarded by `if n_x:` since before this PR, and the structural clause followed the same pattern. Verified against two constructed registers:

| register | count-claims rendered |
|---|---|
| empty | **none** |
| only `look_through` + `not_modelled` | `1 of 2`, `1 of 2` — both live; no false authored/proxy/constructed clause |

**What WAS missing was the pin.** Only the retired `default` case had a `"0 of"` assertion; none of the four table kinds was covered in either direction. Added, parametrised over all four:

- each clause **renders when its kind is present**
- each clause is **suppressed when its kind is absent**
- no `"0 of N"` may render from any register
- suppressing the **counts** does not suppress the **mechanism** — an empty register is a legitimate state and the reader is still owed how the number would be built

Both directions because a suppression that fires *always* is the same defect as one that never fires: the first hides a real population, the second claims an empty one. Same rule `format_drag_exclusion` follows by returning `None` rather than reporting an exclusion of nothing.

**Counts confirmed templated.** M6 hardcodes `n_structural = 1` and `n_authored = 15` and is still killed by `test_the_note_counts_are_templated_not_written`; `test_the_note_counts_reconcile_to_the_row_count` asserts every row is claimed by exactly one clause. That is the #285 reconciliation, and this sentence has exactly the shape that was wrong there.

## Ladder, re-run on `a37263f`

| | |
|---|---|
| Collection | **2105** selected (2145 − 40 deselected) |
| Guards **UP** | 13F / 2065P / 27S = **2105**, reconciles |
| Guards **DOWN** | 13F / 2065P / 27S |
| Names diffed | `diff` empty — **identical across guard states**, and identical to the expected 13 |
| Books | `demo.db` identical across 13 tables; `tracker.db` sha256 unchanged |
| Mutation proof | 12 mutants, re-run, all still KILLED by a named victim |

Issue #244, opened 2026-08-17
reports.py fabricates pct_int = 50 on the exception path: N/A to the reader, a Moderate stance to the allocation logic


MattOrefice:
Filed against #204's missing-marker-surface question, because fixing this **changes rendered PDF output** and therefore needs that decision first.

## The defect

`src/reports.py:1055-1060`:

```python
try:
    _raw_pct = percentile(cape_s, cape_val)
    pct_str  = f"{_raw_pct:.0f}th"
    pct_int  = int(round(_raw_pct))
except Exception:
    pct_str  = "N/A"      # what the reader sees      — honest
    pct_int  = 50         # what feeds the allocation — fabricated
```

`pct_int` flows into `_cape_regime(pct_int)` at `:400`, which maps a percentile to **(regime label, allocation stance)** on bands `>=80 Elevated / >=60 Above-average / >=40 Moderate / else Below-average`. A fabricated 50 lands squarely in **Moderate**.

So one PDF section prints **"N/A"** for the percentile and, from the same failed computation, a **Moderate regime label and allocation stance** — with nothing marking the second as unfounded. The reader is told the input is unavailable and handed the conclusion anyway.

Note the irony recorded at `:395-397`: that comment exists because of a prior same-PDF contradiction, where the executive summary and the Macro section drew opposite conclusions from one CAPE percentile. The single-source fix worked — both sections now agree. They can now agree **on a fabricated number.**

## It is the reachable path, and it does not run through `percentile`

`get_cape_series` (`src/shiller.py:197`) **raises** `FileNotFoundError` when the committed CSV is missing — it does not return an empty series. So:

| state | route | what the reader gets |
|---|---|---|
| CSV missing / unreadable | **raises** → this `except` | "N/A" + a fabricated Moderate stance |
| CSV parseable but all-NaN | `percentile` → 50.0 | a confident "50th" + Moderate stance |

The first is far more reachable — a fresh clone before restore, a deleted file, a partial checkout — and on it **`percentile` is never called.**

## Correcting the sequencing argument that put the Group 2 work first

The argument for doing the `percentile` work before any PDF policy was: *fixing the shared function improves the PDF with no PDF decision required.* **That was true of the narrow path and false of this one.** Fixing `percentile` converts the all-NaN case from a confident "50th" into an honest "N/A" — real, but rare. It does nothing for the missing-file case, which is this issue, and which cannot be fixed without deciding what the PDF renders where a stance used to be.

[...]

PR #301, opened 2026-09-23
The mapping check says when it could not check

MattOrefice:
`unmapped_holdings` returned `{}` from its two broad excepts, which is the same value as a clean book. So the #224 startup notice said nothing when the check had died. This applies #251's shape: the failure path carries its reason.

## The contract

`unmapped_holdings()` returns `UnmappedCheck(disposition, findings, reason, detail)`:

| disposition | when | notice |
|---|---|---|
| `checked` | holdings and `securities` both read | the existing unmapped notice, or none if `findings` is empty |
| `no_holdings` | no dated positions CSV (a first run) | none |
| `could_not_check` | the CSV could not be parsed, or `securities` could not be read | **new**, below |

It is still warn-never-raise, because `app.py` calls it unwrapped at import.

## What the reader sees

Rendered through `app.py` with AppTest, as an `st.warning`:

> **Holdings mapping could not be verified.** At startup, the newest positions file, `Portfolio_Positions_Sep-23-2026.csv`, could not be parsed, so the app cannot say whether every held symbol is mapped in `data/seed/securities_household.csv`. **This is not a finding that everything is mapped:** if any symbol is not, the **Asset Location** and **Household View** pages will fail.
>
> Reason: `KeyError: 1 account number(s) in the export are not present in the account map (private/account_map.json). Add them (mapped to a pseudonym) and re-run; rows are never imported with a raw number.`

- **The verb is per path.** The positions file "could not be **parsed**": the likeliest failure is the account-map step inside the parser, on a file that was read fine, and "read" sends an operator to permissions and encoding first. The `securities` table "could not be **read**", because it was.
- **The reason goes on its own line.** The parser's messages are full sentences, and spliced into the notice they broke it.
- **No raw account number.** The parser's own messages never include one. A test pins that the notice doesn't add one back.

## Verification

- **Render check (AppTest on `app.py`, bootstrap write pass skipped because tracker.db is read-only):**
  - The real book, unpatched: 13 text elements, **no notice**, 0 exceptions. The real check returns `checked` with nothing unmapped.
  - `could_not_check`, produced by the real function on a scratch book whose account isn't in the map: 14 elements, **exactly one** notice element, the `st.warning` above, 0 exceptions.
  - `tracker.db` sha256 unchanged.
- **Tests:** `tests/test_unmapped_holdings_notice.py` gives 21 passed. The bootstrap-related test files give 65 passed plus #177.
  - The two tests that pinned the silent `{}` were migrated to the new contract, not deleted.
- **Mutation proof: 7 mutants, all killed.**

| mutant | killed by |
|---|---|
| M1: parse failure returns a clean `checked` result | rendered-notice test + return-value test |
| M1b: parse failure returns a literal `{}` (the old code) | same |
| M2: `securities` failure returns a clean result | same |
| **M3: return value correct, notice branch deleted** | **only the rendered-notice tests; both return-value tests stay green** |
| M4: exception detail dropped | rendered-notice test |
| M5: parse path says "read" again | rendered-notice test |
| M6: reason spliced back into the sentence | rendered-notice test |

M3 is the separation that matters, the same as #244's M1: a correct return value with a silent notice is the defect, and only a test on the rendered text can see it.


MattOrefice:
Tip is now `1ef769c`. CI's account-number guard rejected the nine-digit literal my new test used as a fake raw account number. The test now uses a non-numeric sentinel; what it checks is whether the account string is echoed, not what it looks like. CI green on `1ef769c5962a35d11722e734226eff18e2b2da19` (run 35886138157): 1973 passed, 0 failed, both PII guards OK. Mutation proof re-run: all 7 killed.

Issue #302, opened 2026-09-23
19 personal-mode render tests depend on the date again: 7-day price look-back vs a cache #269 stopped the suite advancing


MattOrefice:
The personal-mode suite's failure count depends on the date again. **19 render tests** pass while `tracker.db`'s newest price is within 7 days of today, and fail after that. On 2026-09-23 at `7cf5a3a` they fail. At `797d956` (2026-08-23) they passed. CLAUDE.md says "the count no longer depends on the date"; that was true of #241's instance and is false of this one.

## Mechanism (measured)

- The render fixtures copy `data/tracker.db` and block the network.
- `get_sleeve_weights_on_date` looks back **7 days** for a price (`look_back = date - timedelta(days=7)`).
- `tracker.db`'s newest price is **2026-08-19**.

Measured read-only with the network blocked, `sleeve_weights_with_coverage(d)`:

| d | resolved |
|---|---|
| 2026-08-22 | 10 of 10 |
| 2026-08-26 | 10 of 10 |
| **2026-08-27** | **0 of 10** |
| 2026-09-23 | 0 of 10 |

With nothing resolved, `frontier_served` is `None`, and every banner and coverage page renders "No committed price data".

The 19, by file: `test_asof_banner_render` 4, `test_coverage_markers_render` 8, `test_empty_book_render` 4, `test_risk_render` 3. They are attributed through this mechanism; I didn't re-run each one individually against a fresh cache.

## Why the cache stopped advancing: the mechanism is VERIFIED, the history is INFERRED

These are two different claims with different evidence.

**Verified, the mechanism:** the suite still fetches prices, and since #269 those writes land in a per-session copy that is discarded. A leftover redirect copy from one of today's suite runs (`%TEMP%/tracked_db_redirect_f5vlcj49/tracker.db`) holds **484 price rows newer than the real book, through 2026-09-21**. The real `tracker.db` stays at 2026-08-19.

**Inferred, the history:** that suite runs were what kept `tracker.db` fresh before #269. The evidence is circumstantial:
- #269 merged on 2026-08-19, the same date as the book's newest price;
- the #241 record says guards-down personal runs used to advance the cache.

The book's write history is not observable, so this is a date match plus a prior record, not a demonstration. If it holds, #269 did what it was meant to do (the suite no longer moves the book) and in doing so removed what had kept these 19 tests inside their 7-day window.

## Today's live render did not fetch prices (measured)

[...]

PR #311, opened 2026-09-23
get_connection() closes on exit; the redirect cleanup reports leaks

### LINKED ISSUE #307: get_connection() never closes: 97 sites hold connections until GC; surfaced as tracker.db copies left in TEMP
The conftest write redirect (#269) makes a per-session copy of each tracked book under `%TEMP%/tracked_db_redirect_*`. `pytest_unconfigure` removes it with `shutil.rmtree(_redirect_dir, ignore_errors=True)`. **On this machine that cleanup is failing silently.**

Found 2026-09-23:
- **167** leftover `tracked_db_redirect_*` directories.
- **163** of them are empty, all from 2026-08-31.
- **4**, from today's suite runs, each still hold a **16 MB copy of the personal `tracker.db`**: the real household book, outside the repo, in a temp directory.
- One of those copies also holds 484 price rows fetched during the run (that is what verified #302's cause).

## Why it matters

- **Personal data persists outside its intended location**, and outside the gitignore/guard regime that governs `data/tracker.db`. Nothing reports it, because `ignore_errors=True` swallows the failure. That is the same silent-success shape as the check #301 fixes.
- **It accumulates**: one copy per session that touches the personal book.

## Likely cause (unverified)

On Windows, `rmtree` cannot delete a file that still has an open handle. A sqlite connection left open at unconfigure time would make the file undeletable, and `ignore_errors` then hides the failure. The empty directories suggest partial success on other days.

## Not fixed here

The fix is to close or track connections before cleanup, or to fail loud (report which files could not be removed) instead of `ignore_errors=True`. **I have not deleted the leftovers.** They are the owner's data, and one of them is #302's evidence.

MattOrefice:
Closes #307.

## The defect, at its real scope

sqlite3's connection context manager commits or rolls back and **never closes**. So every `with get_connection() as conn:` site (**97**: `src/` 49, `tests/` 32, `pages/` 12, `scripts/` 4) held its connection open until garbage collection. That includes the running app.

It surfaced in the suite. On Windows an open handle blocks deleting a file, so the redirect's per-session copies couldn't be removed. `rmtree(ignore_errors=True)` swallowed the error, and copies of the personal book piled up in `%TEMP%`.

## The fix

- **`get_connection()` passes `factory=_ClosingConnection`.** Its `__exit__` runs the normal commit or rollback, **then closes**. Only `__exit__` changes, so the one explicit-`close()` caller works as before.
- **The redirect cleanup reports what it could not remove.** It runs at terminal-summary time and prints, in red, every path it couldn't remove and the error, where `ignore_errors=True` used to hide it. `pytest_unconfigure` writes to stderr as a backstop. **Nothing already in `%TEMP%` is deleted.**

## Why it's safe: census before the change (details on #307)

- An AST pass over all 97 sites looked for any use of a connection after its block: the connection itself, a cursor from it, a lazy value returned or yielded, or a closure capturing it. It found **0**.
- 53 uses of values from helpers passed the connection were **checked by hand; all are eager** (`read_sql_query` frames, a `.fetchall()` list).
- **The tool was verified before its zero was trusted.** A planted file with one instance of each category: each was caught exactly once, and neither correct pattern was flagged. The first pass flagged 6 sites; all 6 were a later block **rebinding `conn`**. **The tool was fixed, not the findings.**
- **Direct `sqlite3.connect` sites in `src/`: 6, not the 4 first counted.** That count missed `src/seed/`. **All 6 close in a `finally`**; none leaks, so none changed. (Test-side direct connects, e.g. `tests/test_seed.py`'s helper, are outside `get_connection` and outside this PR's scope. The new cleanup report is what would surface one that leaks a tracked-book copy.)

## Verification

- **Before and after on `tests/test_attribution.py`:** 8 open connections and an undeletable redirect dir before; **none and none** after.
- **Full suite, local personal mode, guards up:** **31F / 2069P / 27S = 2127** = `--collect-only`. The failure names are **identical** to the 31 in the 2026-09-23 baseline (#302/#303/#304/#305, #177, #187). No cleanup report was printed, and the `%TEMP%` leftover count was unchanged at 171. demo.db fingerprint identical; tracker.db sha256 unchanged.
- **`tests/test_connection_closes.py`, 8 tests:** closed after the block; a clean block commits before closing; a failing block rolls back and closes; the explicit-`close()` caller works; `row_factory` and foreign keys survive; cleanup returns nothing when removal succeeds; a forced removal failure is reported with its path; and, **on Windows only**, a real open handle is reported.
- **Mutation proof: 4 mutants, all killed.**

| mutant | killed by |
|---|---|
| M1: no closing factory (the old connection) | closed-after-block, rollback-and-close |
| M2: close **before** commit | commits-before-closing |
| M3: commit even on an exception | rollback-and-close |
| M4: `ignore_errors=True` restored | forced-failure test (all platforms) + real-handle test (Windows) |


MattOrefice:
Merged as `590ce0f`. Post-merge proofs: `origin/main..HEAD` empty on main; merge tree `cebd93c3…` equals `7776dfb^{tree}`, the tip CI passed on (run 35892114739, 1980 passed / 0 failed); diff is `src/db.py`, `tests/conftest.py` and `tests/test_connection_closes.py`, with no tracked data; demo.db fingerprint identical; tracker.db sha256 unchanged.

**Acceptance test: `%TEMP%` held at 171 `tracked_db_redirect_*` directories across a full guards-up personal-mode suite run** (31F/2069P/27S = 2127). Before this fix, every session that touched a tracked book added to that pile. The 171 existing directories are untouched; that decision is with the owner.

Issue #305, opened 2026-09-23
Household cash-weight test pins a fact about real holdings; the Sep-23 CSV moved it


MattOrefice:
`tests/test_household_aggregation.py::test_sleeve_weight_within_2pp[cash-4.9]` fails at `7cf5a3a` on 2026-09-23: **cash 1.62% against a snapshot of 4.90%**, beyond the ±2pp tolerance.

**Nothing is wrong with the code. The book changed.** The test reads the newest positions CSV, and `Portfolio_Positions_Sep-23-2026.csv` is now the newest. With that file temporarily moved out, so the Aug-10 CSV is newest, the test passes (the CSV was restored and its sha256 verified).

**This is #178's shape: a pin encoding a fact about real holdings.** `ACTUAL_LT_WEIGHTS` is a snapshot of the household's sleeve weights ("Actuals were computed from May-27 data; tolerance allows small data drift"). Real holdings are expected to move. Every trade, deployment or cash sweep is a potential red here, and the red says nothing about the aggregation code the test exists to protect.

#178 replaced a row-count pin with a two-sided invariant. The same move applies: assert what the aggregation must get right whatever the holdings are. For example:
- weights sum to 100%;
- each sleeve's weight equals its dollar value over the household total;
- a known look-through split is applied;
- or pin against a frozen fixture book, not the live CSV.

Other `ACTUAL_LT_WEIGHTS` sleeves pass today only because they happen to sit within 2pp.

Not fixed here.

MattOrefice:
Another live pin with #305's shape, found while building #309: `test_rendered_prose_char_lengths_pinned` fails on **`relocate_loss_side`** (410 vs 408 pinned) at `main` as well as on the #309 branch, verified both sides. Its rendered length includes live figures the Sep-23 CSV moved. It was **hidden**: the test's loop failed first on `deploy_roth_cash` and stopped there. #309 moves the deploy card to a fixed-balance pin, which is what exposes this one. Not changed in #309.

MattOrefice:
**Chained run, 2026-09-23. Two parts, handled differently.**

- **Text-length pin: fixed in #315.** Every card is now pinned at fixed placeholder values, following #312's fixed-balance precedent for the deploy card.
[...]

PR #314, opened 2026-09-23
A zero Roth balance is scored below the floor, not 10/10 Act now

MattOrefice:
Closes #313.

#312 gave the deploy card tiers, but a **measured zero** fell through `deploy_state` to the group's own **10 / Act now**, because `floor_tiers` has no `zero` row. So **$0 outranked $7.93**, which renders with no score in Accepted.

$0 is below the floor. The zero tier now takes the **below-floor row's score and status** (no score, **Accepted**). **Routing only:** the zero state's wording is unchanged ("Nothing to deploy — the Roth's cash sleeve is empty." and `_DEPLOY_ROTH_CASH_ZERO_STATE`), and a test pins that.

**Tests:** the tier boundary case at $0 now expects `(zero, None, accepted)`. A new test checks that $0 and $7.93 share the subheader (no score) and status, and that the zero wording renders verbatim.

**Mutation proof:** reverting the routing (`row_key = tier`) is **killed** by both tests.

**Local guards-up suite:** reported in a comment below once it finishes.

Issue #304, opened 2026-09-23
Flake: test_bf_per_sleeve_returns_are_total_return, second occurrence (6.4 bps vs 5)


MattOrefice:
`tests/test_attribution.py::test_bf_per_sleeve_returns_are_total_return` has now failed intermittently in **two** full personal-mode runs, so it is a filed flake, not an anecdote.

- **Occurrence 1**: recorded in the local baseline notes. One of three runs failed, and the log carried `price fetch failed (connection reset)`. It passed in the other two, in isolation, and in file context.
- **Occurrence 2** (2026-09-23, `7cf5a3a`): failed guards-up at **6.4 bps against a 5 bps tolerance**, with `r_p 40.41%` against `VEA adj_close 40.47%`. The guards-down run of the same tree passed. The log again had fetch failures.

## Is it guard state or the tolerance? Neither, on the evidence

**Guard state is not the variable.** The conftest write redirect (#269) is on in both states, so the test's writes land in a per-session copy of `demo.db` either way. The one-run difference between the two states is what an intermittent failure looks like, not an effect of the read-only bit.

**The tolerance is not too tight.** Measured on a fresh scratch copy of `demo.db`, network up, twice: the **gap is 0.00 bps** (`r_p 40.152%` = `adj 40.152%`). BF and the endpoint lookup see the same VEA bars (07-20 → 09-21 after the fetch). With the network blocked, BF returns an empty frame and the test **skips**; it can only fail online.

**What does vary:** the failing run shows **different values on both sides** (40.41 / 40.47 against 40.152 in isolation). So by the time this test ran, the shared per-session `demo.db` copy held different VEA prices from a fresh copy's. Earlier tests in the session write into that same copy, and which fetches succeed depends on the network. The leading hypothesis is **mixed-vintage `adj_close`**: rows fetched at different moments carry different dividend-adjustment bases, and BF (path-dependent over the window) and the two-endpoint ratio weigh them differently. Unverified.

## If the hypothesis holds, it is not only a test problem

A long-lived price cache that appends `adj_close` from fetches made on different dates stores prices on different adjustment bases. That is exactly what this test's 5 bps would catch in production data.

## To answer it

Capture the session copy's VEA rows at the moment of a failure, and compare their `adj_close` against a single fresh fetch.

MattOrefice:
## Correction: both explanations offered for this were wrong

[...]

PR #315, opened 2026-09-23
Pin authored prose lengths, not lengths against the live book

MattOrefice:
Part of #305: the text-length pin. The cash-weight part is **skipped** as needing a design decision (see the #305 comment).

The rendered-prose length pin rendered every card against the **newest positions CSV**, so it moved whenever the **book** moved:
- It failed on `deploy_roth_cash` at 496 once the Sep-23 CSV set the Roth cash to $7.93.
- `relocate_loss_side`'s **410 vs 408** hid behind that failure, because a loop reports only its first mismatch. It fails on `main` today.

#312 moved the deploy card to a fixed balance. **This applies the same to every card.** Each placeholder renders as the sentinel `<name>`, so a length moves **only when the authored prose changes**.

- `test_authored_prose_lengths_pinned` **collects every drift before asserting**, so one card can't hide another.
- `test_the_authored_pin_does_not_read_the_book` checks that no placeholder is left unresolved and no $ figure leaks in.
- Whether each placeholder **resolves** in the live render is still checked live (`test_every_placeholder_resolves_into_rendered_output`).
- A stale comment pointer to the retired `RENDERED_PROSE_LEN` is updated.
- The per-card edit history that lived in the old pin's comments is in git history.

**Mutation proof:** dropping one word ("harvested") from a card's prose is **killed**.

**Expected failure-set change:** `test_rendered_prose_char_lengths_pinned` leaves the guards-up failure set. It's retired here, and its replacement passes.


MattOrefice:
Local guards-up suite on `233c166e`: 30F / 2101P / 27S = 2158 = collected. Failures diffed by name and first message against the previous main (#314): the only change is `test_rendered_prose_char_lengths_pinned` **leaving** the set (retired here), plus the #304 flake's volatile figures. No redirect-cleanup report; `%TEMP%` 171 → 171; tracker.db unchanged. CI run 35899635438 green (2010 passed / 0 failed).

PR #316, opened 2026-09-23
Stop the Phase 25.2 migration re-adding accounts.account_number

### LINKED ISSUE #306: Every personal start re-adds accounts.account_number; the next connection drops it
**Every personal-mode start re-adds the vestigial `accounts.account_number` column, and the next writable connection drops it again.**

Reproduced on a scratch copy of `data/tracker.db` (2026-09-23):
- The real book has **no** `account_number` column.
- After `bootstrap_personal_db()` on the copy, it has one, all NULL: `count(account_number) = 0`. The migration log says `Added column 'account_number'` and `Ensured unique index 'ux_accounts_account_number'`. The migration is `tools/migrate_accounts_phase25_2.py`, run by the bootstrap's generic `migrate_*` discovery.
- `_auto_migrate` → `_drop_account_number` (`src/db.py`) then drops it on the next writable open, which is why the real book is currently without it.

## Why it matters

- **CLAUDE.md is wrong about it.** Known Issues (the #232 record) says "once a DB has been opened writably the column is gone and the write stops, so it only appears on a DB restored from an older copy, a second machine, or a clone seeded pre-migration." Every personal start re-creates the condition.
- **It is a DDL write on every start.** It contributes to `tracker.db`'s hash moving on renders that change no data (see #302).
- **No PII.** The column is all NULL every time it is re-added. This is churn, not a leak, but it is churn on the one column the account-number purge was about.

## Not fixed here

The likely fix is retiring the add in `migrate_accounts_phase25_2.py`, now that `_drop_account_number` defines the target schema. Whether a fresh DB still needs that migration for anything else has to be checked first.

MattOrefice:
Closes #306.

Bootstrap runs `migrate_accounts_phase25_2` on **every personal start**. It added `accounts.account_number` and its unique index each time, and `_auto_migrate` m4 (`_drop_account_number`) dropped them on the next writable connection: **an add/drop cycle on every boot**, on the column the account-number purge removed. The column was all-NULL each time, so nothing leaked; it was churn.

## Change
- `account_number` and `ux_accounts_account_number` are removed from the migration's add lists; the docstring says why.
- A regression test runs the migration **twice** on a post-purge table. It asserts the column and its index stay **absent**, and that the other four columns and the pseudonym index still land.
- **Verified on a copy of the real book:** a full `bootstrap_personal_db()` no longer adds the column. The real `tracker.db` is untouched.
- **Mutation proof:** restoring the add is **killed**.

## What this does NOT change: the app still needs write access at boot

The chain's brief asked for this PR to note that it "removes the reason the app needs write access on every boot". **Checked against the code, that's not true, so it isn't claimed:**
- The same migration still **drops and re-creates `ux_accounts_pseudonym` on every run**, a DDL write whether or not `account_number` is added.
- The bootstrap's **seed upserts** (`load_household_securities`, `seed_fund_compositions`, `seed_household_accounts`) write on every start.

A read-only `tracker.db` would still fail at startup, at the next DDL write rather than at `ADD COLUMN account_number`. **CLAUDE.md's session-close rule (#310) stands.** Making boot write-free would be a separate change: make the index step conditional, and make the seeds skip when nothing differs.


MattOrefice:
Local guards-up suite on `ed3d352e`: 29F / 2103P / 27S = 2159 = collected. Failures diffed by name and first message against main `0ff7593`: identical except the #304 flake, which passed this run. No redirect-cleanup report; `%TEMP%` 171 → 171; tracker.db unchanged. CI run 35900734555 green (2011 passed / 0 failed).

Issue #200, opened 2026-08-17
Sweep: pure unit tests carrying an ambient DB dependency


MattOrefice:
## The pattern

A unit test calls a product function that lazily resolves something from `data/tracker.db`, without pinning that input. The test then passes or fails according to the state of the machine's database rather than the behaviour under test — and it usually passes, until the data moves.

**Three instances so far, and the third is what makes it a pattern rather than a habit:**

| # | where | how it surfaced |
|---|---|---|
| 1 | #195 `test_control_condition_reports_complete_coverage` | `assert cov.stale_days > 0` encoded "this machine's cache is behind" as a contract; broke when the suite's own networked run brought the cache current |
| 2 | #198 `test_partial_renders_a_marker...` | `AS_OF_FRONTIER = "2026-08-10"` hardcoded; passed alone, failed in-suite after other tests advanced the cache |
| 3 | #199 `test_banner_with_inception_extends_plain_banner_in_order` | **a pre-existing test nobody had touched.** Pinned the full banner string for a fixed `_TODAY`; the four-state banner resolves a frontier from the DB, so the wording changed under it |

The first two were tests written during this workstream. The third was not — it predates it, and it broke for the same reason. That is the evidence that this is latent across the suite rather than a recent habit.

## Why it is worth a sweep

The failure mode is quiet in the direction that matters: these tests **pass** in the state the author happened to be in. A test that reads the DB to decide what to assert is not testing the function, it is testing the function against today's data, and it will keep being green until someone's cache moves — at which point it fails for a reason unrelated to the change in front of them. In this workstream that cost investigation time three separate times, twice with the failure initially read as a product bug.

## What to look for

Not "tests that use the DB" — the live-data tests legitimately do. The target is **tests that look pure**: no fixture DB, no `_live()` helper, no skip guard, calling a function that resolves something lazily. Candidate signals:

- an expectation that is a date, a price, or a count, written as a literal
- a call to anything that reaches `committed_price_frontier`, `get_prices`, `get_holdings_on_date`, `get_inception_date`, or `get_portfolio_account`
- tests outside `tests/render/` and without `_skip_without_personal_inputs`, that nonetheless fail when `data/tracker.db` is absent

A cheap first pass: run the non-render suite with `DB_PATH` pointed at an empty scratch database and see which pure tests fail. Anything that breaks was reading the real one.

## The fix, per instance

[...]

PR #317, opened 2026-09-23
test_backfill_skips_spaxx runs offline and proves it

### LINKED ISSUE #209: test_backfill_skips_spaxx makes an outbound network call on a cache miss
A #200 instance, found while building PR #207's fixture.

`tests/test_drip.py::test_backfill_skips_spaxx` runs `backfill_all_drip_lots` against `minimal_db`, whose `dividends` table is empty. `fetch_distributions` → `get_dividends` → on a cache miss, `fetch_prices(ticker, start, end)` at `src/prices.py:476-480` — an outbound Yahoo call from a test that is not marked `live_data`. It passes either way (the warn-and-skip path is what the test asserts), so the call is invisible in a green run.

Same class as the socket-patching finding: the test appears hermetic and is not. The fix is to intercept at `src.drip.get_dividends` / `src.drip.get_prices` — the names bound into `src.drip`'s namespace by its `from src.prices import ...`, **not** `src.prices.get_prices`, which does not intercept because drip.py holds its own reference. `tests/test_drip_account_identity.py`'s `no_network` fixture does this correctly and can be lifted.

Belongs in #200's sweep rather than as a standalone fix.

MattOrefice:
Closes #209.

The minimal DB's VOO trade sends a cache miss to `fetch_prices`: **an outbound network call from a unit test**. The code swallows a fetch failure, so the test passed whether the network answered or not. **A raising stub wouldn't fail it either**, which I measured: a fetch that raises still gives a pass. So the premise-check's suggested pin ("make `fetch_prices` raise") wouldn't have worked.

## Change
- Distributions and prices are stubbed in `src.drip`'s namespace, using the `no_network` pattern from `tests/test_drip_account_identity.py`.
- Every `fetch_prices` call is **counted and asserted zero**, a pin that survives the code swallowing the error.

## Evidence
- An independent pytest plugin counted `src.prices.fetch_prices` calls: **1 (VOO) before, 0 after.** `tests/test_drip.py`: 30 passed.
- **Mutation proof:** removing the stubs is **killed** by the zero-fetch assertion.

The broader sweep of pure unit tests carrying ambient dependencies (#200) is separate and still open.


MattOrefice:
Local guards-up suite on `6981ec94`: 29F / 2103P / 27S = 2159 = collected; failures identical by name and first message to main `6f341c5`. No redirect-cleanup report; `%TEMP%` redirect dirs 0 → 0; tracker.db unchanged. CI run 35901826740 green (2011 passed / 0 failed).

PR #319, opened 2026-09-23
The Risk page reads risk contribution's own status

### LINKED ISSUE #318: Risk page crashes in personal mode: it infers risk contribution's status from the factor regression
## The crash

`pages/7_Risk.py` takes its insufficient-history branch when the **factor regression** is insufficient. It then reads `run_risk_contribution()["min_obs"]`, **inferring** that risk contribution must be insufficient too ("covariance needs even more history"). But the two results come from different data:

- **The factor regression** runs over the Fama-French factor cache, which **lags**. Its last date is 2026-06-30, and with portfolio inception at 2026-06-09 it has **n = 14 < 30**, on stale and fresh books alike.
- **Risk contribution** runs over **prices**, which grow with the calendar. With a current price cache it's **`ok`, n = 70 ≥ 60**, and an `ok` result has no `min_obs`.

So with current prices and a lagging factor file, the page raises **`KeyError: 'min_obs'`**. That's a real personal-mode crash, not a test artifact. Found by the #302 diagnostic.

## The stale cache was masking it

On the real `tracker.db` as it stands (newest price 2026-08-19), with no fetch, risk contribution is still **insufficient (n = 48)**, so the page doesn't crash. **Fixing the price-cache staleness (#302) first would have made the page crash**: the day the cache is current, risk contribution reaches 60 while the regression stays at 14.

## Fix

**Read the risk-contribution result's own status; never infer it from the factor regression's.** This is the coverage-record rule: each result carries its status, and that's what the page renders from.

Not gated on factor-file freshness: that's a proxy, and it stays wrong if the two results diverge for any other reason.

MattOrefice:
Closes #318.

`pages/7_Risk.py` took its insufficient-history branch when the **factor regression** was insufficient, then read `run_risk_contribution()["min_obs"]`, **inferring** that risk contribution was insufficient too. The two run on different data:
- **The regression** runs on the Fama-French cache, which **lags**: n = 14 < 30, with the factor file ending 2026-06-30.
- **Risk contribution** runs on **prices**, which grow: n = 70 with a current cache.

An `ok` result has no `min_obs`, so the page raised `KeyError`. **The stale price cache was masking it: fixing staleness (#302) first would have crashed the page.**

## Fix
- The risk-contribution section moves into one function, `_render_risk_contribution_section()`, **byte-identical** (115 lines, verified dedented), called from **both** branches. It reads `rc["status"]` itself.
- **Not gated on factor-file freshness.** That's a proxy for the status, and a proxy stays wrong whenever the two diverge for another reason. Each result carries its own status; the page reads that one.

## Tests (CI-runnable, using the **real** `run_risk_contribution` over synthetic sleeve returns)
- **Current prices + lagging factor file** (regression n = 14, risk contribution 70 days): no exception, the factor empty state renders, and **the full risk-contribution section renders** (volatility metric present).
- **Both short** (40 days): both empty states, with risk contribution's stating **its own** n.
- **Migrated, not deleted:** `…caveat_absent_at_full_confidence` rested on "personal mode never reaches this branch", the same inference, and **passed in personal mode only because the page crashed before rendering any text**. It's now `…caveat_tracks_its_own_band`: the caveat renders **iff** risk contribution's own result is low-confidence, both ways.

**Mutation proof:** restoring the inference is **killed on the rendered page** (current prices, lagging factor file → KeyError).

**Expected failure-set change:** the three #318 risk tests (`runs_without_exception`, `behaves_per_band`, `chart_gated_with_table`) leave the guards-up set. `tests/render/test_risk_render.py` passes 14/14 in personal mode and with `TRACKER_MODE=demo`.


MattOrefice:
Local guards-up suite on `2c36f655`: 27F / 2107P / 27S = 2161 = collected. Failures diffed by name and first message against the `761f01a` baseline: the three #318 risk tests **leave** the set, and the #304 flake re-appears (volatile, known). Nothing else moved. No redirect-cleanup report; `%TEMP%` 0 → 0; tracker.db unchanged. CI run 35906374452 green (2013 passed / 0 failed).

PR #320, opened 2026-09-23
Page 11's scope caption states scope, not existence

### LINKED ISSUE #268: Page 11's scope caption is present-tense and implies figures follow, in states where none do
Split out of #260 (fixed by PR #267) rather than folded into it. Filed because it is a wording question, not the defect #260 described, and because leaving it in a merged PR body guarantees rediscovery.

## The caption

`pages/11_Capital_Deployment.py:132-136`, rendered before the data load and before the empty-book branch:

> Current allocation and drift **are** the **Individual Taxable (Self-Directed)** self-directed taxable book. The account each deployment is *logged to* is chosen in the Deploy section below and may differ.

## What #260 established, so this is not re-litigated

The caption was **not** the contradiction, and it needed no edit as part of that fix. It states **scope** — which figures on this page are about — and that is true in every state, including the three empty-book ones. The contradiction was the old message asserting emptiness (*"No holdings found. Seed the database first."*) one element below it. Rewriting the message removed the contradiction, confirmed across all three renders. **Do not "fix" this by reverting to a coupled caption or by gating it on the data.**

## What is left, and it is small

The present tense reads as an existence claim as well as a scope claim. "Current allocation and drift **are** the X book" implies there is an allocation being shown. In the three empty-book states there is not:

- `unpriced` — holdings exist, none priced, figures withheld
- `no_holdings` — nothing open on this date
- `cash_only` — invested value is zero, weights undefined

A reader who reads the caption first meets a promise the next element withdraws. Nothing is *false*; the ordering is just slightly wrong-footed.

## Why it is not obviously worth fixing

Cost is not in the wording, it is in the render order. The caption is emitted at `:132`; the data that would tell it which state it is in is loaded at `:161`, and `_load_data` is `@st.cache_data`-wrapped. Gating or rewording the caption per state means moving the cached load **above** the page chrome, which changes when the expensive call happens relative to the header and is a page restructure — deliberately declined during #267 rather than folded in.

Two cheaper options that avoid the reorder:

1. **Rephrase to pure scope, tense-neutral** — e.g. "Allocation and drift on this page are scoped to the … book." No data dependency, no reorder.
2. **Leave it.** The message immediately below now explains the state explicitly, which is the mitigation that made the contradiction dissolve in the first place.

## Related

- #260 — the empty-book messages, fixed by #267
- #251 — the disclosure shape the new messages follow

MattOrefice:
Closes #268.

The caption renders **before** the data load, so it can't know whether figures follow. "Current allocation and drift **are** the X book" read as an existence claim that the three empty-book states (`unpriced`, `no_holdings`, `cash_only`) then withdrew.

**Now:** "Allocation and drift figures on this page, **when shown**, are for the **X** self-directed taxable book." It's true in every state. The rest of the caption (deployments can be logged to a different account) is unchanged.

**Render test:** `tests/render/test_page11_scope_caption.py` renders page 11 and pins the conditional form. **It fails against the old wording** (verified by reverting the page alone), which is the mutation proof for a wording change.


MattOrefice:
Local guards-up suite on `b6ce88d1`: 26F / 2109P / 27S = 2162 = collected; failures identical by name and first message to main `5447dec` (the #304 flake passed this run). No redirect-cleanup report; `%TEMP%` 0 → 0; tracker.db unchanged. CI run green (2014 passed / 0 failed).

PR #321, opened 2026-09-23
The yield-curve shape sentence claims only the three tenors it reads

### LINKED ISSUE #239: "Normally upward-sloping across all maturities" is asserted from a three-point test
Filed as a **prose-overstatement** class, which the fail-loud workstream has not scoped. It is not a missing-data substitution — it is wrong with *complete* data.

`pages/3_Macro.py:1072-1073`:

```python
if _yc_10y > _yc_2y > _yc_3m:
    _yc_shape = "normally upward-sloping across all maturities"
```

Three points decide a claim about eight. The chart plots 3M, 1Y, 2Y, 5Y, 7Y, 10Y, 20Y, 30Y — the test reads 3M, 2Y, 10Y. A curve with `10Y > 2Y > 3M` and `20Y < 10Y` (a common late-cycle long-end shape) renders as "upward-sloping across all maturities" while the plotted line visibly falls at the back end. The reader is told something the chart in front of them contradicts.

It reaches the reader inside an interpretive caption, next to a Fed-policy reading:

> Current shape at 3M x%, 2Y y%, 10Y z% is **normally upward-sloping across all maturities**, reflecting current Fed policy and market rate expectations.

**Why it is its own class.** Everything else found in the `pages/3_Macro.py` sweep substitutes a value for missing data. This one has all its data and overstates what the data supports — the same shape as a docstring claiming scope its assertion lacks, but in rendered prose. The fix is not a marker; it is either testing the tenors the sentence quantifies over, or narrowing the sentence to the three it actually reads.

**The other four branches are narrower and probably fine** — "humped — 2Y above 10Y", "inverted at the front end", "mildly inverted at the front end, steepening beyond the 2-year" and "flat to mildly upward-sloping" all scope themselves to the points tested. Only the first generalises to "all maturities".

**Not urgent.** The caption is interpretive rather than decision-driving, and the numbers quoted in the same sentence are real. But it is the kind of sentence a reader trusts precisely because it sits beside correct figures.

Found during the #231-adjacent MARKER sweep of pages/3_Macro.py; deliberately not folded into that work.

MattOrefice:
Closes #239.

The shape is decided from **three** tenors (3M, 2Y, 10Y), but the chart plots **eight**. The upward branch said "normally upward-sloping **across all maturities**", and a curve rising 3M → 2Y → 10Y can still dip beyond 10Y (a common late-cycle shape).

**Now:** "upward-sloping **across those three maturities**". The caption already names the three points in the same sentence ("Current shape at 3M x%, 2Y y%, 10Y z% is …"). The other four branches were already scoped to what they test and are unchanged.

**Test:** `tests/test_macro_curve_shape_claim.py` reads the `_yc_shape` assignments through `ast` (the strings that render, not a text search). It pins five branches, none claiming "all maturities", and the upward branch naming its three points. **It fails against the old wording** (2 of 3 red with the page change reverted).


MattOrefice:
Local guards-up suite on `0036f00a`: 26 failed, 2112 passed, 27 skipped, 40 deselected, 29 warnings ; failures diffed by name and first message against the previous main: no change beyond this PR's own tests (the #304 flake is excluded as known-volatile). No redirect-cleanup report; `%TEMP%` redirect dirs 0 → 0; tracker.db unchanged. CI run 35908570748 green on `0036f00af25ba2c2aabb29c5ca89760262f5c6d7` (pull_request).

PR #322, opened 2026-09-23
Document why pages/ is exempt from the import-time DB guard

### LINKED ISSUE #174: pages/9_Correlations.py:24 calls a DB-backed function at module level — decide whether the #165 import guard should extend to pages/
`pages/9_Correlations.py:24` calls a DB-backed function at **module level**:

```python
_SLEEVES: dict[str, list[tuple[str, float]]] = ae.sleeve_benchmarks()
```

`asset_evaluation.sleeve_benchmarks()` → `_derive_sleeve_maps()` → `sleeve_config` → `_strategic_rows()` → `get_connection()`. So importing this page module opens the database, and — until migration 3 is retired — write-attempts it via `_auto_migrate`.

PR #165 made six `src/` constants call-time and added `test_src_module_import_touches_no_db` to keep them that way. That guard enrolls **`src/` only**, so this line is uncovered.

**It is not currently a bug.** Streamlit re-executes a page script top-to-bottom on every rerun, so a page-level call is call-time by construction — the comment at `pages/9_Correlations.py:21-23` says exactly this, and it is correct. The value does not freeze across an in-process migration the way a `src/` module constant would.

**The question this ticket exists to settle** is whether the guard should extend to `pages/`, and the honest answer is that the two directories want *different* guarantees:

- `src/` — importing must not touch the DB. Unconditional; already enforced.
- `pages/` — a page module *is* allowed to read the DB while executing, because executing is what a page does. What it should not do is freeze a DB-derived value where a warm process would keep serving a stale one, which for Streamlit means module-level state that survives reruns (`@st.cache_data`/`@st.cache_resource`-wrapped, or written into `st.session_state`), not plain module-level assignment.

So a copy-paste of the `src/` guard would flag this line wrongly. If `pages/` gets a guard it needs its own predicate. Worth deciding deliberately rather than leaving the asymmetry undocumented — someone will eventually read "the import guard covers the codebase" and believe it.

Related: PR #165, and the `_auto_migrate` placement work (this line is one of the call sites that reaches `src/db.py:236` today).

MattOrefice:
Closes #174.

#174 asked whether the #165 import guard (`test_src_module_import_touches_no_db`) should extend to `pages/`. **Decided: no.** The reason is now written next to the guard in `tests/test_imports.py`:

- The guard exists for modules **other modules import**. A DB read at import runs as a side effect of `import src.x`, at collection time, in every consumer, before any fixture can scope it.
- **A Streamlit page is not imported by anything.** Streamlit *executes* it as a script on every rerun, so its module body **is** its render. `pages/9_Correlations.py`'s module-level `ae.sleeve_benchmarks()` is the page doing its job.
- What *would* be a page defect is a DB value **cached across reruns** (`st.cache_*`, `session_state`) without invalidation. That's a render-staleness question, not an import-side-effect one.

**Comment only; no behaviour change.**

**Side finding, not in this PR's scope:** `_SRC_MODULES` globs `src/*.py` **without recursing**, so `src/seed/` and `src/ingestion/` are outside both the importability test and this guard. It's the same missed-subdirectory shape as #307's census undercount; listed for the owner.


MattOrefice:
Local guards-up suite on `b4482d11`: 26 failed, 2112 passed, 27 skipped, 40 deselected, 29 warnings ; failures diffed by name and first message against the previous main: no change beyond this PR's own tests (the #304 flake is excluded as known-volatile). No redirect-cleanup report; `%TEMP%` redirect dirs 0 → 0; tracker.db unchanged. CI run 35909581581 green on `b4482d11aa400437bab4e5b39b38ad8268c88b2f` (pull_request).

PR #323, opened 2026-09-23
Rename the proxy test to what it checks: transcription at a fixed as-of

### LINKED ISSUE #234: The proxy recomputation test freezes its as_of, so it is a typo check advertised as a recomputation
`tests/test_yield_proxies.py:279` is named `test_entry_values_match_their_proxies_recomputed_from_the_cache` and its docstring says *"Independent right-hand side: the proxy's TTM is recomputed here from the committed dividends/prices tables, never read back from the config."* The independence claim is true. The **currency** claim is not made explicitly and is what a reader takes away.

`tests/test_yield_proxies.py:291`:

```python
as_of = date(2026, 8, 11)
w0 = (as_of - timedelta(days=365)).isoformat()
```

The recomputation window is pinned to the **authoring** date. As the cache accumulates newer dividends and prices, the window stays fixed, so the right-hand side re-derives the same frozen TTM forever. It catches a mis-transcribed digit. It cannot catch the thirteen entries going stale, which is the failure that will actually happen.

## Why this is the green-means-less shape

It belongs with the six forms already recorded, and it is closest to *"a comment claiming scope the assertion beneath it does not have"*: the test's **name** is the overstatement. "Recomputed from the cache" reads as "checked against reality"; what runs is "checked against a reconstruction of the same moment the value was authored." The assertion passes, so the gap is never evaluated.

It has a second layer. `data/tracker.db` is gitignored, so:

```python
if not db.exists() or db.stat().st_size == 0:
    pytest.skip("personal cache absent")
```

**In CI it skips entirely.** So the one test that validates fourteen yield values against data runs only on one machine, and only as a typo check. The `assert checked >= 10` guard is good and prevents an almost-empty local pass — but it does nothing about the skip, which is form 1 on the list.

## How it misled a real decision

This test was cited twice as evidence for an asymmetry that does not exist:

1. In PR #233's body — "those are also dated snapshots, but a recomputation test recomputes them from the price cache, so drift there has something that fails."
2. In #231's staleness comment, which built on it: "a proxy entry can be recomputed against the cache, so drift has something that fails. A constructed entry re-adds against its own declared components — stale components sum correctly forever."

**Both wrong, in the same direction.** Neither entry type ages. And the residual difference runs the other way: `macro_cache` exists in **committed demo.db**, so a construction check could run in CI where this proxy test skips. The constructed entry is the better-covered of the two.

Not fixed here because un-freezing the window is a decision, not a repair: recomputing against *today* makes the test fail whenever the market moves, which is the desired staleness signal but converts a correctness test into a currency test — and it would still skip in CI. Choosing between "recompute at a moving window", "recompute against demo.db so it runs in CI", and "rename the test to what it does" is scope for its own pass.

Related: #231 (the staleness surface), #200 (ambient-DB test reads).

MattOrefice:
Closes #234.

`test_entry_values_match_their_proxies_recomputed_from_the_cache` **read as a currency check**. What it actually does: recompute each proxy's TTM yield at **one frozen as-of** (2026-08-11) and compare the authored value. That verifies the value was **transcribed correctly** from the measurement at the recorded date. It says nothing about whether the value is still **current**.

- **Renamed** to `test_entry_values_match_their_proxies_at_the_shared_as_of`.
- The docstring now says what it checks and what it doesn't.
- The one cross-reference (`tests/test_yield_construction.py`) is updated.

**The as-of stays frozen, deliberately (decided):** a moving window would make the test's outcome depend on the day it runs, which is the date dependence this repo keeps removing.

**Rename and docstring only; no assertion changes.** `tests/test_yield_proxies.py` + `tests/test_yield_construction.py` pass.


MattOrefice:
Local guards-up suite on `c722109f`: 27 failed, 2111 passed, 27 skipped, 40 deselected, 29 warnings ; failures diffed by name and first message against the previous main: no change beyond this PR's own tests (the #304 flake is excluded as known-volatile). No redirect-cleanup report; `%TEMP%` redirect dirs 0 → 0; tracker.db unchanged. CI run 35910531701 green on `c722109fbe00b2225dc2e27d2ef2d92a09bb28e7` (pull_request).

PR #324, opened 2026-09-23
Pin the decision: a drip-only ticker is shown, not filtered

### LINKED ISSUE #208: backfill_all_drip_lots: silent zero when a ticker's only rows in the account are drip lots
Split out of #201 / PR #207, deliberately not fixed there — it is the site-4/site-5 asymmetry, not the scoping question.

`_traded_tickers` (the ticker list) has **no** `lot_source` filter, while the share-timeline query excludes `lot_source='drip'`. So a ticker whose only rows in the scoped account are drip lots enters the backfill loop, finds no non-drip trades, and falls through `src/drip.py`'s `if not trade_rows` branch — historically `results[ticker] = 0` with **no warning at all**, unlike the two neighbouring skip paths which at least `warnings.warn`.

PR #207 makes it visible as `BackfillResult(0, "no_trades", ...)` and it now renders on the operator's table as `0 — SKIPPED: no non-drip trades` under the INCOMPLETE notice. So it is no longer silent, but the underlying asymmetry stands.

Two candidate resolutions:
1. filter the ticker list to non-drip rows, matching the timeline — makes the branch unreachable, at the cost of a ticker whose non-drip trade was deleted no longer being visible at all;
2. leave the asymmetry and keep the named skip, on the grounds that a drip-only ticker is a data-integrity signal worth surfacing rather than filtering away.

(2) is what #207 shipped by default. Deciding between them is a semantics question about whether a drip-only holding is an error or a state.

MattOrefice:
Closes #208.

**Decided: show it. Hiding a real holding is the worse error.** A ticker whose only rows in the account are drip lots stays in the backfill's ticker list and reports the named skip. It renders as `0 — SKIPPED: no non-drip trades` under the INCOMPLETE notice.

That's option 2, **which #207 already shipped by default**, but **nothing pinned it**. The only test naming `no_trades` just enumerates the status vocabulary. So option 1 (filter `_traded_tickers` to non-drip rows, which makes the holding vanish) could have been applied silently.

**Test:** `test_a_drip_only_ticker_is_shown_not_filtered` requires the drip-only ticker to be **in** the results, with status `no_trades`, and the report to render the SKIPPED label.

**Mutation proof:** applying option 1 (`AND lot_source != 'drip'` in `_traded_tickers`) is **killed**.

No behaviour change; this pins the decision.


MattOrefice:
Local guards-up suite on `ed1f9352`: 26 failed, 2113 passed, 27 skipped, 40 deselected, 29 warnings ; failures diffed by name and first message against the previous main: no change beyond this PR's own tests (the #304 flake is excluded as known-volatile). No redirect-cleanup report; `%TEMP%` redirect dirs 0 → 0; tracker.db unchanged. CI run 35911473187 green on `ed1f9352b705d6517a8ba2720a1d997f5695770c` (pull_request).

PR #325, opened 2026-09-23
Correlations discloses the first plotted date, not the first return

### LINKED ISSUE #258: Correlations: declare the window the series was actually computed over, not the intersection start
Split out of #257, which fixed the exclusion set. This is the disclosure question underneath it.

## The gap

On the demo book, with Extended history enabled:

| | |
|---|---|
| returns intersection starts | **2007-12-12** (SCZ inception) |
| `_ext_start` derived from that | `2007-12-12` |
| **plotted correlation series starts** | **2008-03-11** (after the 60-day rolling burn-in) |

`pages/9_Correlations.py:472` derives `_ext_start` from the *returns* intersection, and the checkbox label (`:478`), help text (`:481-485`), and History-constraint caption (`:609-610`) all interpolate it. So the page says **"reaches ~2007"** and **"extends back to 2007-12-12 and the 2008 crisis becomes visible"**, while what a reader actually sees plotted begins three months later.

## Why the three months are the wrong three months

March 2008 is *after* Bear Stearns (March 14-16). The phrasing "the 2008 crisis becomes visible" implies a pre-crisis baseline to compare against — the whole point of reaching back past the crisis is seeing correlations *before* they converged. That baseline is not in the plot. Lehman (September 2008) is, so the claim is not empty, but it is stronger than the data.

The chart's own caption does print the true start (`x_min` from `mean_s.index.min()`, `:509`), so a careful reader can reconcile it. The two statements sit about forty lines apart and disagree.

## Shape

This is #246 exactly — a percentile declaring the window it was actually computed over. Same fix shape: derive the disclosed start from the rendered series, not from the input that precedes a burn-in.

Note the window is user-selectable (30/60/120/252), so the gap scales: at 252 days the plotted start is roughly a year past the stated one.

## Secondary, same file

`_RECENT_CUTOFF = "2010-01-01"` leaves a gap: a sleeve with a 2008-2010 inception constrains the common window past the crisis but is not offered for exclusion, so the Extended-history option would silently under-deliver. No such sleeve exists on either book today. Worth deciding whether the cutoff should be derived from the crisis date it exists to reach rather than set to a round year.

## Not in scope here

Whether to reword the claim or change the computation is the actual decision. Reworking `_ext_start` to mean "first plotted date" is a render-path change and needs local Streamlit verification per CLAUDE.md.

MattOrefice:
Closes #258.

The Extended-history **label**, its **help text**, and the **History-constraint caption** all interpolated `_ext_start`, the start of the **returns** intersection. They promised "the 2008 crisis becomes visible". The rolling correlation can't plot until one window has elapsed, so the plotted line starts **later**: 2008-03-11 at 60 days (after Bear Stearns), and roughly a year later at 252 days.

**Fix, in #246's shape: disclose the start of the rendered series.**
- **History-constraint caption** (renders *after* the series is computed): states the **first plotted point**, `x_min`, the same date the chart caption prints. It reads "The line above starts **YYYY-MM-DD**, one N-day window after its own returns begin — the disclosed start is the first PLOTTED point, not the first return."
- **Checkbox label and help** (render *before* the computation, so they can't know `x_min`): "returns from ~YYYY", and "the plotted line starts one N-day window after the returns begin".
- The "2008 crisis becomes visible" claim is removed.

**Render test**, parametrised over the **default** and **Extended** views: in the "Average over time" view, the History caption's disclosed start must **equal** the chart caption's first plotted date. **It fails against the old wording** (2 of 2 red with the page reverted). A first draft rendered the default heatmap view and **skipped vacuously**; it now selects the view the caption lives in, and both cases run.

**Not in scope, left on #258:** the secondary `_RECENT_CUTOFF = "2010-01-01"` question.


MattOrefice:
Local guards-up suite on `43120f1d`: 26 failed, 2115 passed, 27 skipped, 40 deselected, 29 warnings ; failures diffed by name and first message against the previous main: no change beyond this PR's own tests (the #304 flake is excluded as known-volatile). No redirect-cleanup report; `%TEMP%` redirect dirs 0 → 0; tracker.db unchanged. CI run 35912393696 green on `43120f1d795f405a20422f079a7eb49ddac59e38` (pull_request).

PR #326, opened 2026-09-23
Each freshness banner states which frontier it uses; "1 day behind"

### LINKED ISSUE #265: Two pages disagree about freshness on the same book: 'Live data as of Aug 18' and 'Prices through Aug 17 — 1 days behind'
Surfaced while diagnosing #241, and filed separately because it is a product question rather than a test one.

## Two pages, one book, one moment, contradictory freshness claims

Rendered against the same scratch copy of `tracker.db` with the price cache stamped current, network blocked, control condition:

```
1_SAA.py       "Live data as of August 18, 2026."
8_Research.py  "Prices through August 17, 2026 — 1 days behind."
```

Same database, same render pass, same day. One page says the data is current; the other says it is a day stale.

## Why, and both halves are deliberate

The two pages reach `as_of_live_line` through different frontier definitions, and the difference is exactly one day whenever today has a bar:

| page | frontier source | today included? |
|---|---|---|
| `1_SAA`, `2_Performance` | `coverage.frontier_served` (they pass `coverage=`) | **yes** |
| `8_Research` | `committed_price_frontier()` (bare `as_of_banner()`) | **no** |

Each choice is documented and load-bearing on its own terms:

- `src/coverage.py:128-141` says `frontier_served` must include `as_of_requested`, and warns specifically against copying the exclusion rule: doing so *"reports None precisely when every price is fully current, i.e. blanks the disclosure exactly when coverage is best."*
- `src/holdings.py:224-233` says `committed_price_frontier` takes `MAX(price_date)` **strictly before today**, so a partial mid-session bar cannot pose as a settled frontier.

Neither docstring mentions the other. Both are right about the question they answer. The contradiction is in the gap between them.

## Consequence

`as_of_live_line`'s **state 1 is unreachable on any page without a coverage record**, because that path's frontier can never equal today. So `8_Research` renders *"N days behind"* on a perfectly fresh cache, permanently, while the coverage pages render *"Live data as of \<today\>"* from the same data.

A reader moving between pages sees two claims that cannot both be simple truths. The likely reading is that one page is broken.

## NOT established

Whether this is intended. It is a consequence of two separately-correct decisions, and it is entirely possible the answer is "yes, and the pages mean different things by 'as of'". I have not found anywhere that decides it, and neither docstring acknowledges the interaction.

If it is intended, the fix is disclosure — the two sentences need to say *which* question they answer, so "1 days behind" does not read as staleness when it means "settled through". If it is not, one definition should feed both.

## Not in scope for #241

#241 is the test-side consequence and is being fixed on its own terms: the banner contract says the two renders are identical in state 1, and the current test asserts they differ. That fix does not touch this, and the per-page asymmetry above is precisely what its derivation has to encode.

## Reproduction

Scratch copy of `tracker.db`, `DB_PATH` repointed, network blocked at the session layer, then stamp the cache current:

```sql
INSERT OR REPLACE INTO prices (ticker, price_date, close, adj_close)
SELECT p.ticker, <today>, p.close, p.adj_close
FROM prices p
JOIN (SELECT ticker, MAX(price_date) mx FROM prices GROUP BY ticker) m
  ON p.ticker = m.ticker AND p.price_date = m.mx;
```

Then render `1_SAA.py` and `8_Research.py` and compare their `st.caption` banners.

MattOrefice:
Closes #265.

Two pages on one book read "Live data as of Aug 18" and "Prices through Aug 17 — 1 days behind". **Both are right; they answer different questions:**
- a coverage record's `frontier_served` is **"how current is what THIS PAGE served"**, and can include today's bar;
- the committed frontier is **"the latest SETTLED close every holding has"**, strictly before today.

**Decided: keep both, and have each banner say which it means.** `as_of_live_line` (which composes every page's banner) now writes:
- `Prices through Aug 16, 2026 (as served to this page) — 2 days behind.` when a coverage record is passed (SAA, Performance);
- `Prices through Aug 17, 2026 (settled closes) — 1 day behind.` otherwise.

**"1 days behind" is now "1 day behind".** State 1 ("Live data as of …") is unchanged: it's only reachable when what was served is complete and current, which needs no qualifier.

**Tests:** `tests/test_asof_frontier_basis.py` pins both bases, the singular and plural, and state 1. **3 of 4 fail against the old text.** The existing banner tokens (`"Prices through"`) still match.


MattOrefice:
Local guards-up suite on `a113f8e4`: 26 failed, 2119 passed, 27 skipped, 40 deselected, 29 warnings ; failures diffed by name and first message against the previous main: no change beyond this PR's own tests (the #304 flake is excluded as known-volatile). No redirect-cleanup report; `%TEMP%` redirect dirs 0 → 0; tracker.db unchanged. CI run 35913351626 green on `a113f8e41b860ea9713b284f18b30364760541ce` (pull_request).

PR #327, opened 2026-09-23
Quarter snapshots lock benchmark constituents and disclose what they miss

### LINKED ISSUE #204: capture_quarter_snapshot silently fails on two tickers that can never be fetched, and reports success
Split out of #193, which bundled three defects. #193 is the **basis** claim (adj_close served as `close`). This is the **immutability** claim, and its sharper half is a defect no recapture can fix.

## The two that can never be captured

`_get_all_snapshot_tickers` (`cache.py:65-74`) unions `securities.ticker` with `asset_classes.benchmark_ticker` — and that second column holds **composite benchmark specifications, not tickers**:

```
is_valid_ticker('VNQ (60%) + DBC (40%)') -> False
is_valid_ticker('VNQ+DBC')               -> False
```

`fetch_prices` validates before requesting and raises. The capture loop swallows it:

```python
    for ticker in tickers:
        try:
            df = _prices_module.get_prices(ticker, inception_str, end_str)
            frames[ticker] = df["adj_close"]
        except Exception:
            pass                          # cache.py:118-119
```

**Every capture, past and future, silently fails on those two and reports success.** A swallow-and-continue inside the function whose only purpose is to freeze a known set of prices — the exact class the census exists to remove, in the one place the repo explicitly promises immutability. A recapture does not fix it: the resolver is asking for something that is not a ticker, and the `except: pass` guarantees nobody finds out.

`sleeve_config.parse_benchmark_spec` already exists to decompose `'VNQ (60%) + DBC (40%)'` into its components, and raises on a malformed spec rather than guessing.

## And six real tickers are missing, which a recapture does fix

Measured — all three stored snapshots hold **20 columns** against **28** resolving tickers, the same eight absent from each:

```
2025Q1 / 2025Q2 / 2026Q1: stored 20, resolver wants 28
absent: AVDV, AVIV, EFV, IDHQ, IQLT, SCZ, 'VNQ (60%) + DBC (40%)', 'VNQ+DBC'
```

AVDV, AVIV, EFV, IQLT, IDHQ and SCZ are in `securities` now and were not when the snapshots were taken.

## Why either half makes the cover claim false

Anything absent from the snapshot falls through to **live** prices:

```python
        if ticker not in snap_df.columns:
            return original_get_prices(ticker, start_date, end_date)      # :153
        ...
        if filtered.empty:
            return original_get_prices(ticker, start_date, end)           # :162-163
```

So a PDF regenerated for a locked quarter prices six real sleeve holdings **today** under a cover reading "Prices locked May 9, 2026", and those figures move on every regeneration. "Prices locked" is a claim about immutability; for those tickers there is none.

## Not in #193

Different claim (immutability vs basis), different fix (the resolver and the capture loop vs the reader). #193's A+B+deletion does not address it: a fresh capture with both bases still stores 20 of 28 columns and still swallows the two malformed specs.

## Suggested shape

- **The capture loop must not swallow.** A ticker it cannot fetch is either an error that fails the capture or a recorded gap that travels with the snapshot — the same choice `PriceCoverage` makes for the live path, and a snapshot is the one place a coverage record is trivially storable alongside the data.
- `_get_all_snapshot_tickers` should decompose composite specs via `parse_benchmark_spec` rather than passing them to a fetcher.
- The cover line should state coverage, not only vintage.

Found in the read-only census (verified slice); every number re-measured directly. Related: #193 (basis), #188 (census).

MattOrefice:
Closes #204.

`capture_quarter_snapshot` fetched `asset_classes.benchmark_ticker` values **raw**. So a composite spec like `VNQ (60%) + DBC (40%)` was requested *as a ticker* and could never succeed, and every failure went to `except Exception: pass`. A ticker absent from the snapshot is **read live** by `snapshot_price_context`, so a report under a **"Prices locked"** cover was partly unlocked, with nothing saying so.

## Decided, and built
- **Composite specs are not tickers.** `_get_all_snapshot_tickers` expands every benchmark spec to its **constituents** with `parse_benchmark_spec`. An unparseable spec is recorded as a gap, never dropped.
- **An unfetchable ticker is a disclosed gap.** Capture stores `(ticker, reason)` (an exception, or no data through the quarter end) and **one bad ticker does not fail the quarter**.
- **Carried end to end:** `SnapshotFrames.gaps` (a defaulted field, so the existing tuple shape and legacy snapshots are unaffected) → stored in the payload → restored by `get_quarter_snapshot` → passed to the template.
- **Disclosed on the PDF cover**, under "Prices locked …": *"Except N series that could not be fetched when prices were locked (X, Y): figures that use them read current prices, not locked ones."* That's accurate because the reader falls through to live prices for them. Nothing renders when there are no gaps.

## Tests (`tests/test_quarter_snapshot_gaps.py`, CI-runnable on a scratch DB)
- Composites expand to constituents; no `+` spec reaches the ticker list; SPAXX is excluded; an unparseable spec becomes a gap.
- One raising ticker: the quarter is still captured, the ticker is a stored gap, and the gap round-trips through the reader.
- The cover disclosure renders from **the real template's cover lines**, sliced from the shipped file (the full template needs the whole report context), and is absent with no gaps.

## Mutation proof, 4 mutants, all killed
| mutant | killed by |
|---|---|
| spec not expanded | expansion test, capture test |
| silent `except: pass` restored | capture test |
| reader drops the gaps | capture test (round-trip) |
| cover disclosure removed | cover test |

`tests/test_cache*.py` + `tests/test_reports.py`: 70 passed with the change.


MattOrefice:
Local guards-up suite on `ad619e32`: 26 failed, 2122 passed, 27 skipped, 40 deselected, 29 warnings ; failures diffed by name and first message against the previous main: no change beyond this PR's own tests (the #304 flake is excluded as known-volatile). No redirect-cleanup report; `%TEMP%` redirect dirs 0 → 0; tracker.db unchanged. CI run 35914320220 green on `ad619e325f618d77ad9147042cc6e1b9697f0126` (pull_request).

PR #328, opened 2026-09-23
Pin look-through weights on a frozen book, not the live household

### LINKED ISSUE #305: Household cash-weight test pins a fact about real holdings; the Sep-23 CSV moved it

MattOrefice:
Closes #305 (its cash-weight half; the text-length half landed in #315).

The household aggregation pin held the **live** book's look-through weights at ±2pp, a fact about real holdings. It failed on cash (**1.62% vs 4.90%**) the day the Sep-23 positions CSV moved the book, and a red said nothing about the aggregation it existed to protect.

## Frozen book, following #315's precedent
Three positions, one fund decomposed through a stated composition. **Every expected weight is worked by hand in the test:**

| holding | how | → sleeve | $ |
|---|---|---|---|
| VOO $6,000 | direct | US Large Core | 6,000 |
| SPAXX $1,000 | direct cash | cash | 1,000 |
| TDF $3,000 | 50 / 30 / 20 look-through | US Large Core / Core Fixed Income / cash | 1,500 / 900 / 600 |

Household $10,000 → **US Large Core 75%, Core Fixed Income 9%, cash 16%**.

**Not a second copy of the look-through.** The expected values are this book's arithmetic, not a general algorithm. A second test applies **#178's rule** (it mustn't be satisfiable by emptiness): the `as_held` weights must be non-empty, sum to 100, and **differ**, so an undecomposed book can't pass.

## Mutation proof: the look-through itself, broken in `src/household.py`
| mutant | result |
|---|---|
| composition weight ignored (equal split) | **killed** |
| whole fund assigned to its first sleeve | **killed** |

**Expected failure-set change:** `test_sleeve_weight_within_2pp[cash-4.9]` (and its nine siblings) leave the suite; the frozen-book tests replace them and run in CI.


MattOrefice:
Local guards-up suite on `155b9d5a`: 25 failed, 2115 passed, 27 skipped, 40 deselected, 29 warnings ; failures diffed by name and first message against the previous main: no change beyond this PR's own tests (the #304 flake is excluded as known-volatile). No redirect-cleanup report; `%TEMP%` redirect dirs 0 → 0; tracker.db unchanged. CI run 35915261228 green on `155b9d5acc75b56f3b329be52ab39cf23cc4979b` (pull_request).

PR #329, opened 2026-09-23
Capture the attribution frontier at collection finish, not at import

### LINKED ISSUE #166: tests/test_attribution.py:491 reads the DB at pytest collection (module-level _HOLDINGS_FRONTIER)
Same defect class as the import-time constants fixed in PR #165, but in a test module: `_HOLDINGS_FRONTIER = _holdings_common_frontier()` at tests/test_attribution.py:491 calls `get_holdings_on_date(today, account_id=1)` at module import, i.e. during pytest collection — opening a connection (and triggering first-connection `_auto_migrate`, a write-attempt) before any test runs. Surfaced by running whole-suite `--collect-only` under a sqlite3.connect trap: collection errors on exactly this line; with the file excluded, collection of the remaining 1,485 tests is clean.

Note: the glob-based import guard added in PR #165 (`test_src_module_import_touches_no_db`) will NOT catch this — it enrolls `src/` modules (plus the render conftest) only. Test modules are outside its parametrization, so this instance needs either its own fix (make the frontier a fixture or lazy) or an extension of the guard to test-module collection.

Found and deliberately left during PR #165 (scope ceiling: report, don't fold in).

MattOrefice:
Closes #166.

`tests/test_attribution.py` computed `_HOLDINGS_FRONTIER` at **module level**, so importing it (which pytest does during **collection**) read the DB. The capture was **deliberately early**: it must happen **before any test's fetch can advance the cache** and move the anchor. So the obvious fix, computing it lazily at first use, would have changed *what* it captures.

## Decided: capture at session start, after collection, before any test runs
- **`pytest_collection_finish`** is the hook that satisfies both constraints: collection is complete, so no import-time DB read, and no test has run yet. It lives in `tests/conftest.py`, since hooks don't register from test modules, and calls the module's `_capture_holdings_frontier()` once.
- The two consuming tests read it through `_captured_frontier()`, which **fails loudly** if the hook didn't run instead of silently computing the value late.
- The module constant is now `None` until captured, and its comment says where and why.

## Tests
- **New:** `test_attribution_test_module_import_touches_no_db` imports `tests/test_attribution.py` under the #165 DB trap (a fresh interpreter whose `sqlite3.connect` raises).
- **Mutation proof:** restoring the import-time capture is **killed** by it.
- The two consuming tests need the personal book, so they run in the local guards-up suite, not in CI.


MattOrefice:
Second mutation (needs the personal book, so run locally): removing the `pytest_collection_finish` capture is **killed**. `test_identity_bf_sum_reconciles_to_stage2` and `test_bf_reconciles_when_wall_clock_past_price_frontier[0,1,3]` fail on `_captured_frontier()`'s assertion, loudly, instead of silently computing the frontier late. Unmutated: `tests/test_attribution.py` 26 passed, 10 skipped.

MattOrefice:
Local guards-up suite on `86c7e793`: 25 failed, 2116 passed, 27 skipped, 40 deselected, 29 warnings ; failures diffed by name and first message against the previous main: no change beyond this PR's own tests (the #304 flake is excluded as known-volatile). No redirect-cleanup report; `%TEMP%` redirect dirs 0 → 0; tracker.db unchanged. CI run 35916226913 green on `86c7e79381d76948fab67d3c3eb6a9f86125e35e` (pull_request).

MattOrefice:
Note, post-merge: `pytest --collect-only` now takes about 6.5s instead of 1.8s, because `pytest_collection_finish` also runs under `--collect-only` and performs the frontier capture (one DB read). That's still **after** collection, so the collection-time property holds; it's just no longer free for a pure collect. Harmless; recorded so the timing change isn't a mystery.

PR #330, opened 2026-09-23
The live-data job opens one issue when it fails, instead of a badge

### LINKED ISSUE #259: A badge is a signal that decays into decoration: live-data was red for 28 days with no push signal
Filed from the #257 diagnostic. Stating the finding; not proposing to decide it now.

## What happened

`test_extended_history_reaches_2008_live` failed on every scheduled `live-data` run from 2026-07-22 through 2026-08-18 — **28 consecutive runs**. It was found by reading a run log during an unrelated state check, not by any signal.

It was not invisible. `README.md:5` carries a linked badge:

```
[![Live Data Tests](https://github.com/MattOrefice/investment-tracker/actions/workflows/live-data.yml/badge.svg)](...)
```

That badge rendered **red on the front page of the repository for 28 days**.

## The actual finding

A badge is a *pull* signal on a surface nobody re-reads. The README front page is read when onboarding and essentially never again by the person who wrote it — so the badge is load-bearing for exactly the audience that doesn't need it, and decorative for the one that does. Its failure mode is silence, and silence is what a green badge also looks like from memory.

Note the asymmetry with the CI badge one line earlier: CI failures surface through PR checks, which block a merge. Nobody relies on the CI badge *as* the signal. The live-data badge has no such backstop — it is the only in-repo signal, and it is the weakest kind.

## What exists today

Across all three workflows (`ci.yml`, `keep-awake.yml`, `live-data.yml`): no `if: failure()` step, no issue creation, no webhook, no notification of any kind.

GitHub's default email to the cron's last modifier is the only push signal, and it is platform behavior — not repo configuration, not visible to review, and evidently not sufficient.

## What a push signal would cost

Roughly, so the decision has numbers attached:

**Option A — open/update an issue on failure.** ~12-15 lines of YAML in a `if: failure()` step using `actions/github-script` or `gh issue`. No new secret (`GITHUB_TOKEN` suffices, needs `issues: write`). The real design question is dedupe: naive creation yields 28 issues rather than one. Search-by-title-then-comment is the usual pattern and is where the complexity actually sits.

**Option B — reuse the badge, add a staleness assertion.** A test asserting the last live-data run on the default branch succeeded. Turns a pull signal into a gate, but couples PR CI to an external workflow's state, which is its own problem.

**Option C — accept it, and record the acceptance.** Rewrite the README line so the badge does not read as monitoring: state that live-data is checked on demand and name who checks it and when. Costs nothing and removes a false assurance, which is most of the harm.

**Option D — webhook to a channel actually read.** Needs a secret and a destination; lowest latency, highest setup.

My read: Option A with dedupe, or Option C. The failure mode being fixed is *nobody learns*, and only A changes that. C is honest about not fixing it, which beats a badge implying otherwise.

## Related

- #257 — the 28-day failure itself
- #258 — the window-disclosure question from the same diagnostic

MattOrefice:
Closes #259.

The README live-data badge was **red for 28 days** and nothing pushed that to anyone. A badge nobody re-reads is decoration.

## A push signal, deduplicated
- On **failure**, the workflow keeps exactly **one open issue**. It's found by the `live-data-failure` **label** rather than by title, which is fragile. The **first** failing run **creates** it; **every later** failing run **comments** on it; it stays open until someone closes it after the job recovers.
- The job gets `permissions: issues: write` (plus `contents: read` for checkout). The label is created idempotently (`gh label create … --force`).
- It runs for scheduled and dispatched runs alike.

## Proof by dispatch, not by reading the YAML
A `workflow_dispatch` input, **`simulate_failure`**, fails the job on purpose so the step can be exercised without waiting for a real outage. **Dispatch results are in the comment below:** run 1 must create the issue, run 2 must comment on the same issue (no second issue). The verification issue is then closed.

**Not added:** auto-closing the issue when the job goes green again. That's easy to add, but it wasn't asked for, and a human closing it is the acknowledgement.


MattOrefice:
## Dispatch proof (on tip `55ef57ce5765550389d70b7ce069d2e02ca9b338`)
- open `live-data-failure` issues before: **0**
- dispatch 1, `simulate_failure=true` (run 35917292069): the job failed as intended, and the step **created #331**
- dispatch 2 (run 35917322086): failed, and the step **commented on #331** ("commented on #331" in the log). **No second issue.**
- after: **exactly one** open issue (#331, 1 comment). It was then closed, since the job isn't really failing.

Verified by running it, not by reading the YAML.

MattOrefice:
Local guards-up suite on `55ef57ce`: 25 failed, 2116 passed, 27 skipped, 40 deselected, 29 warnings ; failures diffed by name and first message against the previous main: no change beyond this PR's own tests (the #304 flake is excluded as known-volatile). No redirect-cleanup report; `%TEMP%` redirect dirs 0 → 0; tracker.db unchanged. CI run 35917292581 green on `55ef57ce5765550389d70b7ce069d2e02ca9b338` (pull_request).

PR #335, opened 2026-09-23
The PDF discloses a figure it could not compute, where it would have gone

### LINKED ISSUE #248: Three except: pass sites feed rendered figures — classify each against its template guard
Deliberately out of scope for the failure-paths PR, which fixed only the two sites that feed a **disclosure channel**. These three feed a **rendered figure**, so whether a swallowed failure discloses depends on each template guard — and that has not been checked per site.

```
src/reports.py:507    try: cape_val = current_cape(); cape_pct = percentile(get_cape_series(), cape_val)
src/reports.py:911    try: fi_result = regress_fi_sleeve(inception, end_date)
src/reports.py:1019   try: bf_df = brinson_fachler_period(...); bhb_...
```

Each leaves a `None` or an absent key on failure. The question per site is the same: does the template render an explicit unavailable state, or does the figure silently disappear, or does a downstream default fill it? The Macro section is known to render honest `N/A` cells, so at least one of these may already be fine — but "may already be fine" is what needs checking rather than asserting.

`:507` is the interesting one: it feeds the executive summary's CAPE sentence, and #244 already covers a *different* CAPE fabrication in the same area (`pct_int = 50` at `:1058-1060`). Whether these are one fix or two depends on whether the exec sentence and the Macro stance share a path.

**Note for whoever picks this up:** the failure-paths PR's test `test_neither_gap_check_swallows_its_failure_any_more` contains an inverse assertion that `:507` is **still an unfixed bare swallow**. Fixing it will turn that assertion red on purpose, with a message telling you to widen the test rather than delete it. That is the intended handoff, not a break.

MattOrefice:
Closes #248.

Three report sites caught every exception with `except Exception: pass`, and then **silently omitted** the dependent sentence. A computation that *raised* read exactly like one with nothing to say.

## One pattern, three sites
`not_computed_sentence(subject, exc, consequence)` produces:

> *{subject} could not be computed for this report ({ExceptionType}), so {consequence}.*

It follows the PDF rule: **what** couldn't be computed, **why** (the exception type, as the existing *Coverage checks incomplete* disclosure does; the message goes to the log), and **what was done instead**. It **never instructs** the reader, and a test pins that. Each site now logs the exception and appends the sentence **exactly where the missing sentence would have rendered**. A working computation adds nothing.

## The three rendered sentences
Rendered by the tests, which force each computation to raise:

1. **Executive summary, CAPE reading:**
   > The CAPE valuation reading could not be computed for this report (Exception), so this summary states no valuation regime.
2. **Factor section, fixed-income note:**
   > The fixed-income sleeve's factor regression could not be computed for this report (RuntimeError), so this section carries no fixed-income note.
3. **Benchmark section, Brinson-Fachler cross-reference:**
   > The Brinson-Fachler attribution cross-reference could not be computed for this report (RuntimeError), so the intercept is shown without the attribution drivers behind it.

(The exception type is whatever was actually raised; `Exception` and `RuntimeError` are what the tests force.)

## Tests (`tests/test_report_not_computed.py`)
- **The pattern:** exact form, and no instruction words.
- **CAPE:** the executive summary built with the reading failing contains the disclosure; built with it working, there's no disclosure and the "CAPE stands at 30.0x" sentence renders.
- **FI and BF:** the real factor and benchmark sections, **built on the demo book so CI runs them**, with the one computation forced to raise. The disclosure is in the section's prose. **Contrast:** built normally, neither section carries any "could not be computed" sentence.

## Mutation proof
| mutant | killed by |
|---|---|
| CAPE disclosure not appended | CAPE disclosure test |
| FI disclosure not appended | FI disclosure test |
| cross-reference disclosure not appended | BF disclosure test |

`tests/test_reports.py` + the new file: 59 passed. The only `except Exception: pass` left in `src/reports.py` is the PDF-renderer fallback (WeasyPrint → xhtml2pdf), which feeds no figure.

**Not verified:** a full generated PDF. The sentences are verified in the section data the template renders sentence by sentence (`{% for sentence in … %}`).


MattOrefice:
Local guards-up suite on `24c13776`: 25 failed, 2133 passed, 27 skipped, 40 deselected, 29 warnings ; failures diffed by name and first message against the previous main: no change beyond this PR's own tests (the #304 flake is excluded as known-volatile). No redirect-cleanup report; `%TEMP%` redirect dirs 0 → 0; tracker.db unchanged. CI run 35921948085 green on `24c13776047ebc3f5a14a732e862b6b90aa7a9d5` (pull_request).

PR #336, opened 2026-09-23
Attribution runs from inception when the window predates it, and says so

### LINKED ISSUE #303: Four of #187's twelve now pass with nothing changed
#187 lists twelve render tests that fail in personal mode. At `7cf5a3a` on 2026-09-23, **four of them pass**, and nothing in the repo was changed to make them pass:

- `test_two_stage_reconciliation_uses_price_series`
- `test_two_stage_attribution_section_renders`
- `test_stage_reconciliation_gap_caption_appears_under_gap`
- `test_naive_gap_notice_and_unavailable_tiles_under_total_gap`

All in `tests/render/test_performance_render.py`. The fifth #187 member in that file, `test_period_returns_no_suppression_on_full_history`, still fails, along with the 4 factor-profile and 3 trade-log members. So #187's twelve are **8** today.

**An unexplained pass is as much a finding as an unexplained failure.** The likeliest explanation is the same date dependence as #302 running the other way: with no holding priced inside the 7-day look-back, the Performance page may take a different branch. That branch could satisfy these four assertions for the wrong reason, or not reach the code that failed them.

Not investigated. To answer it: run these four against a fixture copy of `tracker.db` with prices carried forward to today (as `test_asof_banner_render._scratch("current")` does), and see whether they fail again. If they do, they pass today only because the page is rendering a degraded state, and a green here means nothing.

Measured in the 2026-09-23 baseline: 31F/2057P/27S guards-up, WeasyPrint 70.0, `tracker.db` newest price 2026-08-19.

### LINKED ISSUE #334: Attribution renders nothing, with no reason, when the selected window starts before inception
Found by #303's diagnosis, and it's a **product** defect as well as a test one.

On page 2's Two-Stage Attribution, a trailing window (1M / 3M / 1Y) is `anchor_end − N days .. anchor_end`. When that start falls **before the portfolio's inception (2026-06-09)**, `brinson_fachler_period` returns **0 rows** and the section renders **nothing**: no attribution and no reason. Measured: 3M 2026-05-21..08-19 gives 0 rows; 2026-06-09..08-19 gives 9. A reader choosing **1Y** today sees an empty section with no explanation, and will until 2027-06-09.

**Fix (decided):** clamp the window's start to inception and **disclose** it: *attribution from inception, 2026-06-09, because the selected window starts before it.* That's the same shape as #246's percentile-window captions: state the window actually used, derived from what was computed.

To check as part of the fix: whether clamping also removes the four #187 performance tests' calendar dependence (#303).

MattOrefice:
When the attribution window started before the portfolio's **2026-06-09 inception**, `brinson_fachler_period` computed **zero rows** and the section rendered **nothing**. A reader choosing *1 Year* saw no attribution and no reason why. This was found as #303's test anomaly, and it's a product defect as well as a test one.

## The change
`clamped_period_bounds(period, anchor_end, inception)` returns `(start, end, clamped)`. It's `period_bounds` with the start clamped to inception, plus whether a clamp happened. `_load_attribution` computes over the clamped window, and the page states it where the section begins:

> **Attribution from inception, June 9, 2026**, because the selected 1 Year window would start September 22, 2025, before the portfolio existed. The figures below cover inception to {end}.

This follows the shape of #246's percentile-window captions. The disclosure comes from **the same clamp the computation uses**, so the two can't disagree.

**Why clamping is consistent here:** the other side of the reconciliation already begins at inception. That's the value series `pv` and the naive benchmark, both through `_benchmark_period_return`. The attribution window now agrees with them.

**What's unchanged:** the returns **table**. There, a pre-inception window is still *suppressed*, not clamped (`period_window_predates_inception`), because a "1 Year" return must not describe a months-old book. The docstring says so.

## Tests (`tests/test_attribution_window_clamp.py`)
- **Two-sided at inception:** a 3M window anchored 2026-09-06 starts one day before inception and is clamped. Anchored 2026-09-07, it starts exactly at inception and is not clamped.
- **Since inception is never clamped; 1Y is.**
- **Live page, personal mode:** selecting 1Y shows the disclosure, and the two-stage section renders. This test skips in CI, where there's no personal book.

## Mutation proof
| mutant | killed by |
|---|---|
| never clamp (`return start, end, False`) | 2 of the pure tests |
| off by one (`start <= inc`) | the two-sided test at 2026-09-07 |
| page computes unclamped (`period_bounds` back in `_load_attribution`) | the live test + the rewritten #303 test |
| page drops the disclosure (`if False:`) | the live test + the rewritten #303 test |

## #303's four tests
**Answer: the clamp removes the calendar dependence for all four, and one of them needed rewriting to show it.**

The page's anchor is `_C = last_settled_price_date(...)`. The default 3M window predates inception whenever `_C` is before **2026-09-07**. Today, with the network available, `_C` is 2026-09-22, so 3M isn't clamped and all four pass either way. That is #303's "pass with nothing changed": the calendar moved past 2026-09-07. I pinned `_C` to tracker.db's own frontier, 2026-08-19, and ran the four on each tree:

| anchor | before this PR | after the clamp alone | after the test rewrite |
|---|---|---|---|
| 2026-09-22 | 4 pass | 4 pass | 4 pass |
| 2026-08-19 | **1 pass, 3 fail** (section empty) | 3 pass, 1 fail | **4 pass** |

The one that still failed after the clamp was `test_two_stage_attribution_section_renders`, and it failed **correctly**. It asserted that Stage 1 differs between 3M and SI. A clamped 3M window *is* the since-inception window, so the two figures were equal (`+188 bps` both). The assertion was only true once the book was more than three months old.

The test now **derives which contract applies** from the page's own anchor. When 3M is clamped, the tiles must agree **and** the inception disclosure must be shown. When it isn't, the tiles must differ **and** there's no disclosure. Both anchors pass, and both page mutants above fail it. (`test_two_stage_reconciliation_uses_price_series` passed at the stale anchor even before this PR. That's a separate question I didn't pursue.)

This leaves #187 with the returns-table test (`test_period_returns_no_suppression_on_full_history`, whose suppression is by design), the 4 factor-profile tests and the 3 trade-log tests. #303 is answered here.

Closes #334. Closes #303.


MattOrefice:
Note on the failure diff: 7 of the 25 first-failure messages differ from the previous main only in a traceback line number inside `pages/2_Performance.py` (line 390 → 392, the lines this PR adds above it). Normalising that page's line numbers leaves 0 differing messages; names identical.

MattOrefice:
Local guards-up suite on `4c7fcfde`: 25 failed, 2138 passed, 27 skipped, 40 deselected, 29 warnings ; failures diffed by name and first message against the previous main: no change beyond this PR's own tests (the #304 flake is excluded as known-volatile). No redirect-cleanup report; `%TEMP%` redirect dirs 0 → 0; tracker.db unchanged. CI run 35923177323 green on `4c7fcfde3cbd31cce5b3a9dedc752ee3a6b285b8` (pull_request).

PR #337, opened 2026-09-23
A blend's yield basis names its least-verified component

### LINKED ISSUE #297: look_through names a mechanism, not a provenance: a blend over mixed underlyings reports neither kind
Named in #289's Q5 design and left out of scope there. Filing it because #289's PR made the surrounding vocabulary precise enough that this gap now stands out.

## The gap

`_assumed_yield_with_source` returns `look_through` for a blend regardless of what its underlyings declare. After #289 the other five bases each name a provenance — `proxy`, `constructed`, `authored`, `structural`, `not_modelled` — and `look_through` names a *mechanism*. It says the value was decomposed; it says nothing about what it was decomposed **into**.

So a blend over one proxied sleeve and one authored sleeve renders as neither:

```
GAOSX   multi_asset   look_through   (underlyings: some measured, some judged)
```

A reader who has just been told that `(authored)` marks an unmeasured judgement gets no such marking on a row whose value is **partly** one — and the weighted average of a measured 3.97% and an authored 6.00% is not a measurement.

## Why it is the same shape as #294, one level down

**#294**: a sleeve is one label doing several jobs — exposure, yield keying, tax character — and each new axis is a chance for the label to be right for one and wrong for another.

**Here**: a *basis* is one label doing two jobs — how the value was **derived** (look-through) and what it was derived **from** (a mix of kinds). `look_through` answers the first and silently drops the second, exactly as `table` answered "where is it stored" and dropped "how was it arrived at" before #289.

The fix #289 used generalises: **a basis should name the weakest provenance in the chain, not the mechanism at the top of it.**

## Options

1. **Propagate the weakest.** A blend whose underlyings are all proxied reports something like `look_through(proxy)`; one with any authored component reports `look_through(authored)`. Truthful and cheap; makes the render more crowded.
2. **Mark only when mixed or weak.** Unmarked when every underlying is declared-and-measured, marked otherwise — the same "declared or not" rule #289 settled on for the flat case.
3. **Leave it and disclose in the note.** Cheapest; the note already partitions rows by basis, so it could state what the look-through rows rest on without changing the column.

Option 2 is the most consistent with what just shipped, but this is a decision rather than an obvious fix.

## What it does not affect

**No rendered figure is wrong.** The look-through arithmetic is correct and #210 already refuses a blend with any `not_modelled` component, so the worst case is a blend of *declared* kinds whose provenance is under-described rather than misdescribed. This is a disclosure gap, not a numeric one.

Scope note: on the current book the only blend rows are `multi_asset` (GAOSX) and `target_date`, and GAOSX already resolves to `not_modelled` because it holds 5% commodities — so **zero rows exercise the mixed case today**. That is an argument for filing rather than fixing, and an argument against assuming it will stay that way.

## Related

- **#289** — split `table` for exactly this reason at the flat level
- **#294** — the same one-label-many-jobs shape at the sleeve level
- **#210** — established that a blend with an unsizable component refuses whole rather than normalising

MattOrefice:
`look_through` named a **mechanism**, not a **provenance** (#297). A blend over a proxied sleeve and an authored one rendered as plain *look-through*. So a judgement was hidden behind a label that reads as mechanical, and a weighted average of a measurement and a judgement read as a measurement.

## The change
- **`blend_weakest_basis(symbol, compositions_df)`** returns the least-verified basis among a blend's components. It resolves each component with **the same `_assumed_yield_with_source` the yield uses**, so the two can't disagree about a component.
- The register carries it as **`yield_basis_weakest`** on every look-through row, and `None` elsewhere.
- Page 14 renders the column through **`assumed_yield_cells`**, one helper, so the page can't drop the component while the formatter knows it:
  > 2.48% (look-through; least-verified component: authored)
- **The ordering is a stated choice** (`BASIS_VERIFICATION_ORDER`, with a comment giving the reasoning):
  - **authored:** a judgement.
  - **proxy:** measured, but the benchmark is chosen.
  - **constructed:** arithmetic on published series.
  - **structural:** a fact about the asset.

This implements #297's option 1 (always name the weakest). The issue leaned towards option 2 (mark only when mixed or weak); the chain instruction settled it as option 1.

## Rendered check (page 14, personal mode, scratch copy of tracker.db)
**No rendered cell changes on the current book.** Before and after, the Assumed Yield column is identical: 33 cells (15 authored, 4 not modelled, 1 structural, 13 unmarked). **None is a look-through row**, which is exactly #297's own scope note ("zero rows exercise the mixed case today": GAOSX resolves to *not modelled*). The change is inert today and fires on the first blend of declared kinds.

## Tests (`tests/test_blend_weakest_basis.py`)
- An authored component is named even at 5% weight.
- A measured-only blend names its weakest measurement.
- No usable composition names nothing.
- The ordering puts the judgement first.
- The formatter's string.
- **Wiring:** a blend held in taxable reaches the **register** with `yield_basis_weakest == "authored"`, and a non-blend row carries none.
- **Wiring:** `assumed_yield_cells` carries it to the rendered cell.

## Mutation proof
| mutant | killed by |
|---|---|
| `max` for `min` (names the most-verified) | 3 tests |
| register never populates the column | the register wiring test |
| cell helper drops the component | the rendered-column test |

Closes #297.


MattOrefice:
Local guards-up suite on `4ff6aecb`: 25 failed, 2145 passed, 27 skipped, 40 deselected, 29 warnings ; failures diffed by name and first message against the previous main: no change beyond this PR's own tests (the #304 flake is excluded as known-volatile). No redirect-cleanup report; `%TEMP%` redirect dirs 0 → 0; tracker.db unchanged. CI run 35924199688 green on `4ff6aecb6d1bcf5cc1267b5f177d976dcd7e3240` (pull_request).

PR #341, opened 2026-09-24
A leaked connection to a tracked-book copy fails deterministically

### LINKED ISSUE #340: test_seed's _conn() leaks a connection to the redirect copy (#307 class), visible only when run alone
`tests/test_seed.py`'s `_conn()` returns a bare `sqlite3.connect(str(_DEMO_DB))`, and every caller uses `with _conn() as conn:`. That commits or rolls back but **never closes**. The write-mode open of `data/demo.db` is redirected to a per-session copy (#269), and the unclosed connection holds it at session end.

**Measured 2026-09-23 at `a15b993`, demo mode:** `python -m pytest tests/test_seed.py::test_drip_increases_portfolio_value` prints `tracked-DB redirect cleanup FAILED — a connection leaked (#307)` (WinError 32 on the copy) and leaves one `%TEMP%/tracked_db_redirect_*` directory holding a ~28 MB demo.db. Six such directories exist now, all from targeted runs today. None were deleted.

**Why the full suite reports 0:** garbage collection closes the connection before session teardown in a long run, so the leak only shows in a short one. The cleanup report is right, and this is the case it exists for.

**Fix:** make `_conn` a closing context manager (the `_ClosingConnection` shape `get_connection` uses), or wrap the callers in `contextlib.closing`.

MattOrefice:
## The leak
`tests/test_seed.py`'s `_conn()` returned a bare `sqlite3.connect(...)` used as `with _conn() as conn:`. That **commits but never closes**. The open is a write-mode open of the tracked demo.db, so the suite redirects it to a per-session copy, and the handle kept that copy open. It's now a closing context manager.

The #340 issue named one test; **three** leaked, because every caller of the helper did.

## The detection gap
#311's cleanup reports a leak only if garbage collection hasn't closed the connection by session end. So a short run reported test_seed's leak and **the full suite reported clean**.

**Is there a warning to promote?** No. Python **3.11.9's sqlite3 emits no ResourceWarning** for an unclosed connection (checked: `-W error::ResourceWarning -X dev`, then `gc.collect()`, raises nothing). That warning arrived in 3.13.

**So the redirect detects it itself, deterministically.** It holds every connection it opens by a **strong reference**, tagged with the phase and the test that opened it. Garbage collection therefore can't close a leak before it is checked, and nothing calls `gc.collect()` before a check.

| opened during | checked | on a leak |
|---|---|---|
| a test's **call** phase | right after that test's teardown | that test **errors**, by name, so it lands in JUnit |
| setup or collection (a module- or session-scoped fixture may legitimately hold one across tests) | at session end | the **run fails** (exit 1), with each opener named |

Either way the harness closes the connection after reporting it, so the copy can still be removed. The registry drops closed connections at each check, so it stays small over a full run.

## Evidence
- **Positive control, full suite (the case that used to hide it):** with test_seed's fix reverted, the full suite ends **9F / 2165P / 27S + 3 errors**, the three test_seed tests. Before, it reported clean.
- **With the fix:** full suite **9F / 2165P / 27S, 0 errors**. No other test leaks a redirected connection.
- **Self-tests** in `tests/test_connection_closes.py`:
  - garbage collection can't close a leak before the check;
  - the check reports **and closes** an open connection, and drops closed ones;
  - a setup-phase connection is left for the session-end check.

## Mutation proof
| mutant | killed by |
|---|---|
| redirect doesn't register the connection | the GC test |
| reported but not closed | the report-and-close test |
| per-test check off | the session-end backstop: test_seed's leak still named, and pytest exits 1 |

## %TEMP%
The six leftover `tracked_db_redirect_*` directories, each holding a demo.db copy from targeted runs, are **deleted** as authorised. **0 remain.**

Closes #340.


MattOrefice:
Local guards-up suite on `8fc16134`: 9 failed, 2168 passed, 27 skipped, 40 deselected, 39 warnings ; failures diffed by name and first message against the previous main: no change beyond this PR's own tests (the #304 flake is excluded as known-volatile). No redirect-cleanup report; `%TEMP%` redirect dirs 0 → 0; tracker.db unchanged. CI run 35942433020 green on `8fc1613412d37dd923849ec8dd262fa7c0f4ed87` (pull_request).

PR #342, opened 2026-09-24
Render tests run against a frozen test book, offline, on a pinned date

### LINKED ISSUE #187: 12 render tests are demo-mode-only: the suite cannot be run in personal mode

### LINKED ISSUE #302: 19 personal-mode render tests depend on the date again: 7-day price look-back vs a cache #269 stopped the suite advancing

MattOrefice:
**#302's test half: a frozen test book, and it closes #187.**

## Why
- **#302's render tests copied the owner's `tracker.db`.** They **skipped in CI** (no tracker.db, no positions CSV, no account map), and locally they went red when its price cache fell behind the calendar.
- **#187's three modules rendered whatever book was configured.** They failed in personal mode because the real book takes different branches: an inception too young for some windows, and content the personal render doesn't surface.

## The frozen test book: `tests/fixtures/frozen_book.db`
Built by **`tools/build_frozen_book.py`**:
- **Public origin only.** Built from `data/demo.db`, **never** `tracker.db`. The builder refuses any other source, and `tests/test_frozen_book.py` asserts in CI that its accounts, trades, securities and theses are all rows of demo.db.
- **#304's format.** Raw closes with `adj_close` NULL, plus the dividend record. The builder refuses a source that still stores `adj_close`.
- **Nothing date-keyed.** `macro_cache` rows (keyed on the fetch date) are dropped.
- **Trimmed** to prices and dividends from 2024-01-01; inception is 2025-05-01 and the frontier is 2026-07-20. **2.2 MB**, and **byte-reproducible** (three builds, one sha256: `19c61709…`).
- **It carries the 23 seeded theses**, 18 asset classes and 5 themes.

**Always used through a copy, with the network blocked.** A live fetch would append today's prices and make the book depend on the date, and an in-place open would mutate a committed fixture.

## The date: a frozen book alone is not enough
The first offline run against the frozen book failed 4 Performance tests. The page reported **"Benchmark data unavailable for S&P 500 (SPY) as of 2026-09-23"**: it checks benchmark coverage as of the *real* today, which is two months past the book's frontier. The pages call `date.today()` 79 times (YTD, quarter reportability, benchmark coverage). So against a fixed book, the *same* render changes with the calendar; in 2027, YTD alone would move.

So the fixtures also **pin today to the book's next day, 2026-07-21** (`FROZEN_TODAY` in `tests/conftest.py`):
- It's a `date` subclass whose `today()` returns the pin and whose `isinstance` still accepts real dates (from pandas, sqlite, arithmetic).
- It's patched into `datetime` and into every loaded `src`/`tests`/`pages` module that bound the real `date`, **for the module's duration only**.
- **Restored afterwards, including in any module first imported under the pin.** That module would otherwise have kept the pinned class for the rest of the run; `unpin_leftovers()` handles it, and a test pins it.
- `datetime.now()`'s three uses are not reached offline: the unsettled-bar check (fetch only), snapshot capture time, and Macro's "Last updated" caption. A test asserts **no loaded `src`/`pages` module escapes the pin**.

The Performance page's benchmark-coverage-as-of-today is a real offline date dependence in the **product** too, the same class as #302's product half. I'm recording it here, not fixing it: pinning the clock makes the tests date-independent, and the product question is separate.

## What moved onto the book
| module | before | now |
|---|---|---|
| `test_asof_banner_render`, `test_coverage_markers_render`, `test_empty_book_render` (#302) | copied `tracker.db`; skipped in CI | a copy of the frozen book, clock pinned, **run in CI** |
| `test_performance_render`, `test_factor_profile_render`, `test_trade_log_render` (#187) | the configured book | a copy of the frozen book, offline, clock pinned, in **every** mode |

**Results, both modes:**
- #302's three modules: **41 passed**, personal and demo.
- #187's three: **48 passed**, personal and demo. Against `tracker.db` they were 8 failed / 36 passed / 4 skipped. The 4 skips were "No portfolio data", and they now render.

**#187's eight all pass:**
- 4 factor-profile tests;
- 3 trade-log tests;
- `test_period_returns_no_suppression_on_full_history`.

**#177 is now the suite's only red.**

**Trade-off, stated:** #187's modules no longer render the *live* `demo.db` in CI; they render its frozen copy. What demo.db holds is still asserted by the demo-book tests (`test_unadjusted_storage`'s committed-state checks, `test_seed`, the per-table and golden tests).

## .gitignore
Every `*.db` stays ignored (fail-closed). **One named re-admit:** `!tests/fixtures/frozen_book.db`, never a pattern. A test asserts this file is admitted while `tests/fixtures/another_book.db`, `tests/fixtures/tracker.db` and `data/tracker.db` are still ignored.

## Mutation proof
| mutant | killed by |
|---|---|
| a trade row that isn't demo.db's | the public-origin test |
| a stored `adj_close` | the format test |
| a date-keyed `macro_cache` row | the format test |
| the re-admit widened to `!tests/fixtures/*.db` | the ignore test |
| the clock pin disabled | the pin test **and** 4 Performance render tests (benchmark "unavailable as of" the real today) |
| `unpin_leftovers` a no-op | the late-import test |

CLAUDE.md is updated: the #302 bullet now describes the frozen book and the pin, and #177 is the only deliberate red.

Closes #187. Closes #302.


MattOrefice:
Local guards-up suite on `7123b04f`: 1 failed, 2196 passed, 23 skipped, 40 deselected, 170 warnings ; failures diffed by name and first message against the previous main: no change beyond this PR's own tests (the #304 flake is excluded as known-volatile). No redirect-cleanup report; `%TEMP%` redirect dirs 0 → 0; tracker.db unchanged. CI run 35945201281 green on `7123b04f08ad9ecb9c8af634fb9d015a730016fd` (pull_request).

PR #338, opened 2026-09-23
The price look-back runs from the newest stored price, not the calendar

MattOrefice:
**The product half of #302.** Every "current price" lookup looked back **7 days from the requested date**. When the cache was more than a week behind the calendar and the gap couldn't be fetched (offline, or a failed fetch), every window came back empty. Nothing resolved, and pages rendered **"No committed price data"** for a book that *has* prices, just older ones. Offline use degraded only because the calendar moved.

## The change
`look_back_start(ticker, date)` starts the window **7 days before the newest stored price on or before the date**.
- **Online:** `get_prices` still fetches the gap, exactly as before.
- **Offline:** it serves the stored price, and the coverage record's `frontier_served` states how old that price is. **The banner discloses the lag instead of the page going empty.**
- **Unchanged** for a historical date (the newest stored price is within the week) and for a ticker with no stored price.

It's applied at **all four sites** that open-coded the look-back: both valuations in `holdings.py`, `positioning._portfolio_market_values`, and `tax_lots.get_lot_inventory`.

## Tests (`tests/test_look_back_from_stored.py`, demo book, offline, so CI runs them)
- The look-back anchors on the newest stored price, and a historical date is unchanged.
- **Two months past the newest stored price, offline:**
  - sleeve weights resolve, and `frontier_served` is the stored date;
  - positioning values every position above zero;
  - every lot has a current price.

## Mutation proof
| mutant | killed by |
|---|---|
| `look_back_start` anchors on the date again | the anchor test and the sleeve-weights test |
| positioning site reverts to the calendar look-back | the positioning test |
| tax-lot site reverts to the calendar look-back | the lots test |

## Effect on the personal-mode suite: 16 of #302's render tests now pass
These are the failures #302 attributed to the 7-day look-back against a cache that stopped advancing at 2026-08-19. The four files, run alone at this tip: **55 passed, 0 skipped**. Diffed by name against the previous main, these 16 leave the failure set and nothing enters it:

- `test_asof_banner_render::test_banner_is_the_only_element_that_changes[control-2_Performance.py]`
- `test_asof_banner_render::test_banner_is_the_only_element_that_changes[partial-2_Performance.py]`
- `test_asof_banner_render::test_partial_banner_differs_from_control_banner[1_SAA.py]`
- `test_asof_banner_render::test_partial_banner_differs_from_control_banner[2_Performance.py]`
- `test_coverage_markers_render::test_control_page11_still_assesses_bands`
[...]

PR #344, opened 2026-09-24
The Performance page judges benchmark coverage at its settled frontier

### LINKED ISSUE #343: Performance page judges benchmark coverage as of today, not as of its settled frontier
Found by the frozen test book (#342), which fixed the tests by pinning the clock and left the product as it was.

**The defect.** `pages/2_Performance.py` loads the S&P 500, blended and naive benchmark series through `TODAY` (`_load_benchmarks`, `_load_naive_benchmark`) and **then** clips each to the settled frontier `_C` (`last_settled_price_date`), which the rest of the page anchors on. Coverage is judged inside `src/benchmarks._get_price_series` against the *requested* end: a component with no price within 5 days of it is a gap. So the gap is judged against the calendar, while the page displays data only through `_C`.

**Effect.** Offline, or whenever the cache is more than 5 days behind today, the page warns "Benchmark data unavailable for S&P 500 (SPY) as of <today>" even though it holds benchmark prices through `_C` and shows nothing past it. Measured against the frozen book, whose prices end 2026-07-20, on 2026-09-23 with the network blocked: the warning fires, and 4 Performance render tests failed until the clock was pinned.

**Fix, on #338's precedent** (look back from the newest stored price, not the calendar): judge coverage at `_C`, the newest settled price. Online nothing displayed changes, since every series is already clipped to `_C`.

MattOrefice:
**#343.** The Performance page judged benchmark coverage **as of today**, while showing every benchmark series only **through its settled frontier C**. This is the offline weakness #342 found and didn't fix.

## The defect
`_load_benchmarks` and `_load_naive_benchmark` fetched through `TODAY`, then the page clipped each series to `_C` (`last_settled_price_date`, which the rest of the page anchors on). Coverage is judged in `src/benchmarks._get_price_series` against the requested end: a component with no price within 5 days is a gap. So the gap was judged against the calendar, for data the page never displays.

**Effect:** offline, or whenever the cache is more than 5 days behind, the page warned **"Benchmark data unavailable for S&P 500 (SPY) as of <today>"**, even though it held prices through C.

## Fix, on #338's precedent
The loaders take `end` and are passed `_C`, the newest settled price, not the calendar. All three call sites are covered (the S&P/blended pair, the 60/40 tile series, and the selectable naive baseline).

## Evidence
- **Test** (`tests/render/test_benchmark_coverage_at_frontier.py`): page 2 rendered against the frozen test book (prices end 2026-07-20), offline, with today pinned to two dates. The render must be non-empty, so the check can't pass vacuously.

| today | before the fix | after |
|---|---|---|
| 2026-07-21 (the book's next day; the control, where old and new agree) | passes | passes |
| **2026-09-23 (65 days past the book)** | **fails:** "Benchmark data unavailable for S&P 500 (SPY) as of 2026-09-23" | **passes** |

Both modes.

- **Online, nothing changes:** page 2 rendered on the **personal book, network up**, before and after: **0 of 73 rendered elements differ**. Every series was already clipped to C.

## Mutation proof
| mutant | killed by |
|---|---|
| S&P/blended loader back to "as of today" | the 65-days-later case (the page renders "unavailable") |
| naive loader back to "as of today" | the 65-days-later case |

Closes #343.


MattOrefice:
Local guards-up suite on `e08305fd`: 1 failed, 2198 passed, 23 skipped, 40 deselected, 170 warnings ; failures diffed by name and first message against the previous main: no change beyond this PR's own tests (the #304 flake is excluded as known-volatile). No redirect-cleanup report; `%TEMP%` redirect dirs 0 → 0; tracker.db unchanged. CI run 35947625665 green on `e08305fd1eec80aa6192ec14671ae6d61cd569e0` (pull_request).

Issue #345, opened 2026-09-24
asset_classes has no unique key on (name, parent_id): should it get one?


MattOrefice:
Split out of #288, which asserts the predicates instead of changing the schema.

**The gap.** `asset_classes` has no unique index beyond its `asset_class_id` primary key, and `name` is not unique: **"Real Assets" is two rows in both books**, id 3 (the parent, `parent_id` NULL) and id 13 (the sleeve, `parent_id` 3). Every name-keyed migration in `tools/` disambiguates with `parent_id IS NULL` / `IS NOT NULL`. Dropping that clause widens silently to both rows. The #288 test shows it: `WHERE name = 'Real Assets'` matches 2 rows.

**What #288 did instead.** `tests/test_asset_classes_update_predicates.py` checks every `UPDATE asset_classes` predicate in the repo against both books, and fails on any new statement with no rule. That catches a widened predicate. It can't stop a direct write that duplicates a (name, parent) pair.

**The question: a unique index on `(name, parent_id)`?** It's the schema-level guarantee. SQLite treats NULLs as distinct in a unique index, so two top-level rows with the same name would still be admitted. That's the parent case, so the index would need `COALESCE(parent_id, 0)` or an expression index.

**Why it's its own decision: it's the #283 data-path shape.** It touches a **tracked binary**. Adding an index to `data/demo.db` is a committed binary change that deploys publicly, and it must stay additive and inert (CLAUDE.md, Modes). tracker.db needs the same change applied locally. So it's a data PR, verified per table, not a schema line in `src/db.py`, and it has to confirm neither book holds a duplicate pair first. Both hold none today: the only duplicated name, Real Assets, has distinct `parent_id`s.

MattOrefice:
**Closed with a decision: detection is enough for this repository. No unique index.**

#346 (`tests/test_asset_classes_update_predicates.py`) already verifies every `UPDATE asset_classes` predicate in the repo against both books, and fails any new one that lacks a rule. A unique index would add **prevention on top of detection** at the cost of a schema migration on the committed, publicly deployed `demo.db`: the #283 data-path shape (a tracked binary, a data PR verified per table, additive-and-inert constraints). That cost isn't justified by what it would add here.

**When to revisit: an INSERT into `asset_classes` without an existence guard.** That's the one case the test catches only *after* the bad row exists: a duplicate (name, parent) pair makes a name-keyed predicate match 2 rows, and #346 fails then, not before.

[...]

PR #346, opened 2026-09-24
Every UPDATE asset_classes predicate matches the row it means, and no more

### LINKED ISSUE #288: rowcount == 1 should be a migration convention, not a local choice — and assert the WHERE predicate across tools/ rather than retrofitting six scripts
Split out of #283 so the convention lands where the next migration author meets it, rather than only in one script's docstring.

## The failure has no signal

A migration whose `WHERE` matches more rows than intended **succeeds silently**. sqlite raises nothing, the script prints its happy path, and the suite stays green — because the suite asserts the rows the migration *meant* to change, and those are correct. The extra rows it also flattened are the ones nobody asserts.

Into a **tracked binary** it is worse. `data/demo.db` is committed, `git diff --stat` says nothing useful about a SQLite file, and the damage surfaces in a *rendered* diff, if at all, after the merge.

`tools/migrate_demo_vgit_jurisdiction.py` (#283) is the first migration in `tools/` to guard it:

```python
cur = conn.execute("UPDATE securities SET holding_rationale = ? WHERE ticker = 'VGIT'", (updated,))
if cur.rowcount != 1:
    conn.rollback()
    return 1
```

## Should the six existing scripts be retrofitted? **No — and here is why**

| script | predicate |
|---|---|
| `migrate_demo_intl_rationales` | `WHERE ticker = ?` |
| `migrate_demo_missing_falsifiers` | `WHERE ticker = ?` |
| `migrate_demo_revisit_clause_bold` | `WHERE ticker = ?` |
| `migrate_demo_vea_rationale` | `WHERE ticker = 'VEA'` |
| `migrate_personal_intl_rationales` | `WHERE ticker = ?` (×2) |
| `migrate_personal_vea_rationale` | `WHERE ticker = 'VEA'` |

And the schema:

```sql
CREATE TABLE securities (
    ticker TEXT PRIMARY KEY,
    ...
```

Unique in both books — 27/27 in demo.db, 74/74 in tracker.db, enforced by `sqlite_autoindex_securities_1`.

**So the predicate can match at most one row by construction.** `rowcount` ∈ {0, 1}, and 0 is already caught by every script's pre-read (`if row is None: ABORT`). A retrofit would add an assertion that **cannot fire on the code as written** — a check with no failure mode. That is the shape this workstream has spent weeks removing, and it would mean six edits to shipped scripts that write a tracked binary, for no behaviour change.

## What to do instead: assert the PREDICATE, once

The guarantee comes from the `WHERE` being the primary key, **not from anyone's care**. So pin that, across all of `tools/`:

> every `UPDATE securities` in `tools/migrate_*.py` is keyed on `ticker`

Parse with `ast`/`sqlparse` or match the SQL literal; either is fine at this size. **That check has a real failure mode** — it fires the first time someone writes `WHERE holding_rationale LIKE ...` or `WHERE sleeve_category = ...`, which is exactly the mutant #283 used (M2). It covers the six retroactively **without editing them**, and it covers the seventh before it is written.

New scripts should keep `rowcount == 1` anyway: two lines, and it is the local guard for the case where the predicate is genuinely not a unique key — a future `UPDATE prices`, `UPDATE trades`, or any table without a single-column PK.

## Where the convention should live

Not only in a docstring. Options, in the order I would take them:

1. **The test above** — unmissable, and it is the only form that acts on a script nobody read.
2. A short "writing a data migration" section in `tools/` — the checklist is small: target the DB explicitly (never `get_db_path`), be idempotent, abort on unexpected text rather than overwriting, do a targeted substring swap rather than a whole-string overwrite, key the `WHERE` on a unique column, assert `rowcount`, expose `main()` only so bootstrap does not auto-run it.
3. Registration in `tools/sync_personal_prose.py` for personal-side scripts, which already documents why these are manual.

## Related

- #283 — the migration that introduced the guard, and the mutant that proved it

MattOrefice:
**#288, the real version:** assert the `WHERE` predicate once, across the code, instead of editing shipped migrations that write a tracked binary.

## Why it matters here
`asset_classes` has **no unique index** on `name`, and **"Real Assets" is already two rows in both books**: id 3 (the parent, `parent_id` NULL) and id 13 (the sleeve, `parent_id` 3). Name-keyed migrations disambiguate with `parent_id IS NULL` / `IS NOT NULL`. A predicate that drops that clause **succeeds silently** and flattens both rows.

## Enumerated from the repo root: 9 sites, not 5
Every string literal (and f-string literal part) containing `UPDATE asset_classes`, in every `.py` outside `tests/`, found by parsing:

| file | statements | predicate |
|---|---|---|
| `tools/migrate_saa_phase25.py` | 3 | `name = ? AND parent_id IS NULL`, `… IS NOT NULL`, `rationale LIKE ?` |
| `tools/migrate_saa_phase38a.py` | 2 | `name = ? AND parent_id IS NULL`, `… IS NOT NULL` |
| `tools/migrate_saa_phase39.py` | 3 | `asset_class_id = ?` ×2, `name = ? AND parent_id IS NOT NULL` |
| `src/db.py` (`_auto_migrate`'s Real Assets heal) | 1 | `_REAL_ASSETS_LEGACY_WHERE`, name + parent + benchmark |

## The rules, and why not a flat "exactly one"
Counted today on both books, **no predicate matches more than one row**, so nothing is skipped. But several match **0**, each correctly:
- **"International Developed"** matches 0 in **demo.db** because phase39 renamed it there. It matches 1 in tracker.db, where the intl restructure is held.
- Two of **phase25's rationale patches** match 0 because they're already applied; `REPLACE … WHERE LIKE` is idempotent.
- **db.py's legacy heal** matches 0 because it has already healed.

So the test states the invariant per shape:
- **Name-keyed:** **exactly one** row per name the script lists, on each book. The only allowed zero is a name **a later migration retired in that book**. That exception is derived from phase39's own `OLD_NAME`/`NEW_CORE` and holds only where the new name exists and the old one doesn't.
- **Text-keyed and the conditional heal:** **at most one**.
- **`asset_class_id = ?`:** unique by the schema, so the test checks that the primary key is `asset_class_id`.
- **Completeness:** any `UPDATE asset_classes` without a rule fails the test until someone states which row it means to touch. That covers the next migration before it lands. Rules for statements that no longer exist also fail.

tracker.db is checked locally; it's personal and absent in CI, so that case skips there.

## Mutation proof
| mutant | killed by |
|---|---|
| phase38a's sub-class `UPDATE` loses its `parent_id` clause (edited in the shipped script) | completeness: an unruled predicate |
| the same widening stated as a rule | **data:** `WHERE name = ? ('Real Assets',) matches 2 row(s)` |
| a new `tools/` migration nobody registered (`WHERE parent_id IS NOT NULL`) | completeness |
| the retirement exception removed | demo.db (International Developed matches 0) |
| the exception made unconditional | tracker.db (the name still matches 1) |

## Not in this PR
No schema change. The unique-index question is **#345**. It touches a tracked binary, so it's the #283 data-path shape: a data PR verified per table, not a line in `src/db.py`.

Closes #288.


MattOrefice:
Local guards-up suite on `ee7fbfd0`: 1 failed, 2201 passed, 23 skipped, 40 deselected, 170 warnings ; failures diffed by name and first message against the previous main: no change beyond this PR's own tests (the #304 flake is excluded as known-volatile). No redirect-cleanup report; `%TEMP%` redirect dirs 0 → 0; tracker.db unchanged. CI run 35948461642 green on `ee7fbfd09dd30986e7cc570c1c1f0db30337b488` (pull_request).

PR #348, opened 2026-09-24
seed_saa inserts plainly; its empty-table check is the named sole guard

### LINKED ISSUE #347: seed_saa's INSERT OR IGNORE claims a dedup the schema does not provide; the empty-table check is the only guard
Found while closing #345.

**The trap.** `src/seed_saa.seed()` inserts the SAA taxonomy with `INSERT OR IGNORE`. But `asset_classes` has no unique constraint beyond its `asset_class_id` primary key, which the insert doesn't supply, so **`OR IGNORE` ignores nothing**. It reads as deduplication and provides none.

**The only real guard** is the check at the top of `seed()`: if `asset_classes` already has rows, skip. `seed()` runs from bootstrap on **every personal-mode start**, so that check is what stops the taxonomy being inserted again each time.

**The failure it invites.** An editor who sees `OR IGNORE` will reasonably read the empty-table check as redundant and remove it. The next start then **silently duplicates the whole taxonomy**: every parent and sleeve a second time. Nothing errors. Name-keyed migrations would start matching two rows (#346 would then fail, but only after the duplicates exist).

**Fix:**
1. A plain `INSERT`, with a comment that the empty-table check is the sole guard and why `OR IGNORE` can't help without a constraint. Behaviour is identical; the statement stops claiming a property the schema doesn't provide.
2. A test that seeds one database twice and asserts `asset_classes`' row count and names are unchanged by the second run, so removing the check fails a test instead of duplicating the taxonomy.

**Not a schema change:** #345 stays closed (reasoning in the PR).

MattOrefice:
**#347.** `src/seed_saa.seed()` inserted the SAA taxonomy with `INSERT OR IGNORE`. But `asset_classes` has no unique constraint beyond the primary key these inserts don't supply, so **`OR IGNORE` ignored nothing**. It read as deduplication and provided none.

**The only real guard** is the empty-table check at the top of `seed()`, which runs from bootstrap on **every personal-mode start**. An editor who took `OR IGNORE` at its word could reasonably remove that check, and the next start would silently double the taxonomy.

## The change
1. **Plain `INSERT`** for both statements. A comment above the check names it as **the sole guard against duplication** and says why `OR IGNORE` can't help without a constraint. **Behaviour is identical**; the statement stops claiming a property the schema doesn't provide.
2. **`src/bootstrap.py`'s comment** repeated the claim (`# asset_classes (INSERT OR IGNORE)`). It now says what actually happens: `skips if non-empty; #347`.
3. **`tests/test_seed_saa_idempotent.py`** seeds one fresh database twice. It asserts the first seed wrote the full taxonomy (the premise), and that the second leaves `asset_classes`' row count and names unchanged. That pins the real guard.

## Mutation proof
| mutant | result |
|---|---|
| the empty-table check deleted | **killed:** the second seed doubles the taxonomy, **17 rows → 34** |
| the check deleted **and `OR IGNORE` restored** (the belief this issue names) | **killed the same way:** `OR IGNORE` prevents nothing |

The second row is the issue's claim, demonstrated.

## Why #345 stays closed (also on #347)
A unique index would make `OR IGNORE` honest, but:
- **"Real Assets" legitimately appears twice** under different parents, so the key must be `(name, parent_id)`.
- **SQLite treats NULLs as distinct in a unique index**, so two top-level rows with one name would still be admitted. Catching that needs an expression or partial index.
- Either is a schema migration on the committed, **publicly deployed demo.db**, the #283 data-path shape.

This targeted fix protects **the one path that relies on the missing constraint**, with no schema migration.

Closes #347.


MattOrefice:
Local guards-up suite on `e98da66d`: 1 failed, 2202 passed, 23 skipped, 40 deselected, 170 warnings ; failures diffed by name and first message against the previous main: no change beyond this PR's own tests (the #304 flake is excluded as known-volatile). No redirect-cleanup report; `%TEMP%` redirect dirs 0 → 0; tracker.db unchanged. CI run 35950846168 green on `e98da66d2a84c88541bb90a415e7e0cbdb6e7ce3` (pull_request).

Issue #349, opened 2026-09-24
Stage 2 and Total tiles report deposits as return: the portfolio side is an end-over-start value ratio


MattOrefice:
## Symptom

On the personal book, the Performance page's **"Stage 2: Implementation"** tile shows **+10295 bp** and **"Total: Portfolio vs. 60/40"** shows **+10475 bp**, on a 71-day book whose own time-weighted return is **3.2%** (SI TWR 3.16%).

Measured 2026-09-23 on a scratch copy of the personal book, rendered headless with the network blocked; price-cache frontier 2026-08-19; default attribution window (3M, clamped to inception 2026-06-09).

## Mechanism

`pages/2_Performance.py`, in the two-stage section:

```python
_r_p_ps  = _benchmark_period_return(pv, bf_period)
```

`_benchmark_period_return` divides a series' end value by its start value. That is right for a **benchmark**, a normalised series with no cash flows. It is wrong for the **portfolio**: `pv` is a dollar value series that includes deposits, so new money is reported as return. On the personal book, `pv` roughly doubles over the window because of purchases funded by new money, not market moves.

The page already computes the cash-flow-neutral figure: `period_return("daily", pv, cf, ...)` is the SI TWR it shows (3.16%). Stage 2's portfolio side should use that. `_benchmark_period_return` stays as it is for the benchmark and naive series.

## The page's own check caught it and rendered the figure anyway

The BF summary caption shows `vs. Stage 2: ⚠ -10424.04 bps` (`_bf_reconciled = abs(_bf_s2_gap_bps) < 0.5`), and the tiles above it still show the fabricated figure: an honest warning next to a fabricated number.

## How it became visible

**#334 exposed it.** It made the attribution section render on the default window when that window starts before inception. Before #334 the section computed zero rows and rendered nothing on this book. #334's live test (`test_page2_discloses_and_renders_a_clamped_window_live`) checks only that the section appears, not what its figures say. A fix surfaced a latent defect, like the Risk page.

The defect needs a deposit inside the window, not a clamped window. Any window containing a deposit should be affected; to be measured on both books in a follow-up comment.

`test_identity_bf_sum_reconciles_to_stage2` (tests/test_attribution.py) re-implements the page's call (`_bpr_helper(pv_real, label)`) and pins the BF bridge against it on the demo book. It asserts a copy of the defective expression, so it cannot see the defect, and it has to move with the fix.

## Plan (one PR each)

[...]

PR #353, opened 2026-09-24
The quarterly tile, return chart and absolute return are TWRs

MattOrefice:
> **Stacked on #352.** It uses #352's independent reference, `tests/first_principles.py`. Merge #352 first; this PR's base then retargets to `main`.

## What was wrong

Three more Performance figures divided the portfolio's end value by its start value on `pv`, a series that steps up on every deposit. New money read as return, the #349 defect in three more places.

## The change (`pages/2_Performance.py`, `src/returns.py`)

| Figure | Was | Now |
|---|---|---|
| Quarterly return tile | `_q_ret(pv)`: end / start over the quarter | `twr_daily_linked` over the same slice. The quarterly PDF has always computed it this way. |
| Cumulative-return chart, Portfolio line | `pv / pv.iloc[0]` | `twr_index(pv, cf)`, new in `src/returns.py`: `twr_daily_linked`'s chain kept as a series (`index[-1] − 1` equals it; unit-tested) |
| Reconciliation caption, "absolute" return | `current_val / pv.iloc[0] − 1` | the since-inception TWR. `current_val`, which only this read, is removed. |

The benchmark sides are unchanged: a value ratio is correct for a series with no flows.

**Why the caption is converted, not relabelled.** The brief said to relabel it if it meant value growth, and to skip it if the intent was unclear. The intent is return, and it is documented: `docs/phase_11_diagnostic.md` §2.1 set the absolute return's denominator to `pv.iloc[0]` so that "absolute_return ≡ TWR" holds by construction, and the README lists "TWR equals absolute return for the lump-sum case" among the identities the suite holds. The value ratio kept that identity only until the first deposit. Converting keeps the documented identity on every book.

## Tests

The frozen book has no deposit, so **new `tests/render/test_performance_figures_are_twr.py` plants one** (VOO, 2026-06-01), inside the Q2 2026 window and since inception. It checks each figure against #352's independent reference: positions and flows from the ledger, never `src.holdings` or `src.returns`.

- **Premise:** the deposit lies inside both windows, and moves a value ratio more than 1000 bps off the TWR in each.
- **Quarterly tile** = the quarter's independent TWR, within 0.006 pp. The window comes from `most_recent_reportable_quarter` (the quarter runs from the **prior quarter-end's close**: Q2 = 03-31 → 06-30). My first draft guessed 04-01 and failed by 75 bps, which is why the test now takes the page's own window definition.
- **Chart:** the Portfolio line on the deposit day, 06-30 and the last day each equal the independent TWR since inception, within 1e-6 pp.
- **Caption:** "absolute" equals the independent SI TWR and equals the caption's own TWR figure.
- **`tests/test_returns.py`:** `twr_index` starts at 1.0, ends at `1 + twr_daily_linked`, nets a deposit, and skips a zero prior value exactly as `twr_daily_linked` does.

## Mutation proof (worktree at this commit; each mutant asserts it applied)

| Mutant | Killed by | Message |
|---|---|---|
| quarterly back to `_q_ret(pv)` | `test_quarterly_tile_is_the_quarters_twr` | "Q2 2026 tile reads 37.35%, but the book's TWR over 2026-03-31..2026-06-30 is 11.7338%" |
| chart back to `pv / pv.iloc[0]` | `test_cumulative_chart_is_the_twr_index` | "the chart's Portfolio line reads 65.5412% on 2026-06-01, but the book's TWR since inception to that day is 34.6707%" |
| caption back to the value ratio | `test_absolute_return_caption_is_the_twr` | "absolute return reads 63.9% (TWR 33.3%), but the book's TWR since inception is 33.317%" |

Each mutant is caught only by its own test (1 failed, 3 passed). Control: 4 passed.

## On the owner's book (offline scratch copy, frontier 2026-08-19)

| Figure | main (85b24c3) | this PR |
|---|---|---|
| Caption absolute / TWR | 107.6% / 3.2% | **3.2% / 3.2%** |
| Chart, Portfolio line on 2026-08-19 | 107.575% | **3.159%** |
| Q2 2026 tile | 2.02% | 2.02% (no deposit in Q2; Q3 starts on the deposit day) |

The demo book has no flow after inception, so none of its figures move.

Refs #349.


MattOrefice:
Closing and reopening only to trigger CI: this PR was stacked on #352 (CI runs on PRs into main only), and retargeting to main does not fire the workflow. Head unchanged at 0242a168ffb67132b7cf01ada2664e240c1e47c7.

